PR #179, opened 2020-04-08
feature/REG-74-DID-STORE

JGiter:
### Summary
|             |   |
|-------------|---|
| Description | Proposes claims store interface and its reference implementation based on IPFS |
| Jira issue  | https://energyweb.atlassian.net/browse/REG-74 |

### List of Features
- Added package `did-store-interface` with proposed `IDIDStore` store interface which should implement custom implementations
- Added package `did-ipfs-store` with IPFS-based implementation of `IDIDStore`
- For testing purposes added scripts to start and shutdown IPFS daemon in `test` folder
- `IResolver.read(did)` can read hash of content saved in `serviceEndpoint`
- In `DIDRegistry` class added `publishPublicClaim` and `publishPrivateClaim` methods which adds approved claim to document along with claim hash. These methods also accepts optional arguments which allows to choose hash function implementation. By default `sha256` is used
- Registry accepts `IOperator` instead of `IResolver` because Registry responsible for saving claims
- the `document` member of the `DIDRegistry` now of `IDIDDocumentFull` type instead of `IDIDDocumentFactory` because there is no reason to prefer `IDIDDocumentLite` over `IDIDDocumentFull` 
- Tests in `did-registry` package changed to use claim storage functionality
- Changed Node version to 11.15.0


packages/did-resolver-interface/src/models/resolver.ts
```diff
@@ -58,6 +58,7 @@ export interface IServiceEndpoint {
   description?: string;
   validity?: utils.BigNumber;
   block?: number;
+  hash?: string; // hash of the content located at service endpoint
```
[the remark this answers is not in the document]
JGiter: service endpoint may be the address of a web resource whose contents may vary. Currently only claims are hashed in order to verify their integrity https://github.com/energywebfoundation/ew-did-registry/blob/482a1e6d96682c45ab0d3c7981522874d24e4d0a/packages/claims/src/claimsUser/claimsUser.ts#L311

packages/did-store-interface/src/interface.ts
```diff
@@ -0,0 +1,15 @@
+
+export interface IDidStore {
```
ni3gavhane, line 2: Add method `delete(uri: string)` to the `did-store-interface`. 
For IPFS implementation, you can not delete the content as such but you can display an error "Not supported with IPFS."

JGiter: done

PR #189, opened 2020-04-23
Bug/verify claim fails

drgorb:
### Summary
|             |   |
|-------------|---|
| Description |   |
| Jira issue  | REG-275  |

#### Type of request

### List of Features
- add delegate only once


.eslintrc.json
```diff
@@ -34,6 +34,12 @@
         "int32Hint": false
       }
     ],
+    "no-plusplus": [
+      "error",
+      {
+        "allowForLoopAfterthoughts": true
```
drgorb, line 40: this is to remove linting errors in for loops with i++

package.json
```diff
@@ -93,7 +93,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run lint && npm run compile"
+      "pre-commit": "npm run fix && npm run compile"
```
drgorb, line 96: I noticed a lot of linting errors. If we fix with every commit, we might reduce the occurrences

PR #211, opened 2020-09-25
bug/Incorrect pub key recovery

JGiter:
### Summary
|             |   |
|-------------|---|
| Description | Fixes public key related bugs |
| Jira issue  |   |

#### Type of request

- [x] Bug fix

### List of Fixes
- Fixed public key recovery algorithm
- Private key in claims package optional now, because it can't be calculated from Signer

### DIFF
--- a/packages/claims/src/claims/claims.ts
+++ b/packages/claims/src/claims/claims.ts
@@ -1,4 +1,4 @@
-import { ethers, Signer } from 'ethers';
+import { utils, Signer } from 'ethers';
 import { IDIDDocumentFull } from '@ew-did-registry/did-document';
 import { IJWT, JWT } from '@ew-did-registry/jwt';
 import { IKeys } from '@ew-did-registry/keys';
@@ -6,6 +6,10 @@ import { IDidStore } from '@ew-did-registry/did-store-interface';
 import { DelegateTypes, IPublicKey } from '@ew-did-registry/did-resolver-interface';
 import { IClaims } from '../models';
 import { hashes } from '../utils';
+
+const {
+  arrayify, keccak256, hashMessage, recoverPublicKey,
+} = utils;
 
 /**
  * @class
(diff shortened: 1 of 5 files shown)


packages/claims/src/claims/claims.ts
```diff
@@ -54,10 +58,13 @@ export class Claims implements IClaims {
...
       return this.keys.publicKey;
     }
-    const hash = await ethers.utils.keccak256(await this.signer.getAddress());
-    const sig = await this.signer.signMessage(ethers.utils.arrayify(hash));
+    const hash = keccak256(await this.signer.getAddress());
+    const sig = await this.signer.signMessage(arrayify(hash));
+    const digest = arrayify(hashMessage(arrayify(hash)));
     // eslint-disable-next-line no-return-assign
-    return this.keys.publicKey = Promise.resolve(ethers.utils.recoverPublicKey(hash, sig));
+    const publicKey = `02${recoverPublicKey(digest, sig).slice(4, 68)}`;
```
ni3gavhane, line 65: why is this '02' prefixed?

packages/did-ethr-resolver/src/implementations/operator.ts
```diff
@@ -74,14 +78,15 @@ export class Operator extends Resolver implements IOperator {
...
       return this._keys.publicKey;
     }
-    const hash = await ethers.utils.keccak256(await this.getAddress());
-    const sig = await this._signer.signMessage(ethers.utils.arrayify(hash));
+    const hash = keccak256(await this.getAddress());
+    const sig = await this._signer.signMessage(arrayify(hash));
+    const digest = arrayify(hashMessage(arrayify(hash)));
     // eslint-disable-next-line no-return-assign
-    return this._keys.publicKey = ethers.utils.recoverPublicKey(hash, sig);
+    return this._keys.publicKey = `02${recoverPublicKey(digest, sig).slice(4, 68)}`;
```
ni3gavhane, line 89: same here. 02?

packages/claims/src/claims/claims.ts
JGiter: Its prefix used in public keys. For uncompressed key the prefix is 04, for compressed are - 02, 03

packages/did-ethr-resolver/src/implementations/operator.ts
JGiter: Prefix used to differentiate between compressed and uncompressed key

ni3gavhane (approved):

PR #228, opened 2020-10-16
Feature/myen 193 track identity status

JGiter:
### Summary
|             |   |
|-------------|---|
| Description | Extended proxy model and utility functins |
| Jira issue  | https://energyweb.atlassian.net/browse/MYEN-193 |

#### Type of request

- [x] New feature

### List of Features
- [x] list owned tokens
- [x] list created tokens
- [x] list all tokens
- [x] update metadata uri's individually instead of per baseUri basis
- [x] added tests. They can be used as examples (packages/proxyIdentity/test/proxy.facade.test.ts)


packages/proxyIdentity/test/proxy.test.ts
```diff
@@ -333,17 +332,4 @@ describe('[PROXY IDENTITY PACKAGE/PROXY CONTRACT]', function () {
       expect(await proxy2.owner()).equal('0x0000000000000000000000000000000000000000');
     });
   });
-
```
ni3gavhane, line 336: why we removed this test?

ni3gavhane: we need test for updating URI.

ni3gavhane (approved):

PR #234, opened 2020-10-24
feature/separate proxy and token ownership

JGiter:
### Summary
|             |   |
|-------------|---|
| Description | PR separates ownership of proxy from ownership of token. Proxy owner can be changed while token is permanently owned by the proxy which creates it on deploying. Proxy contract identified by string serial numb. ERC1155 token should be called only through the proxy |
| Jira issue  | https://energyweb.atlassian.net/browse/MYEN-258 |

### List of Features
- [x] approved agent of proxy owner also becomes approved agent of proxy
- [x] overrided `ERC1155.mint` which increments count of created tokens. This counter needed to set id of newly created token
- [x] added `ProxyManager` class which provides utility functions like `allProxies`, `proxiesOwnedBy` and batch proxy operations like `createProxyBatch` and `changeOwnerBatch`
- [ ] `ERC1155.burn` and `ProxyIdentity.burn` - do we need them and what they should do?


tests/integration/proxyIdentity.test.ts
```diff
@@ -121,12 +111,17 @@ describe('Identities shared management with proxies', function () {
     expect(await store.get(claimUrl)).equal(claim);
   });
 
-  it('Installer transfers ownership from OEM to asset owner', async () => {
```
ni3gavhane, line 124: We need this test. can we add it.

ni3gavhane: ok. I think it is covered in directly.

ni3gavhane (approved):

PR #244, opened 2020-11-02
bug/create proxies txransactions have same nonce

JGiter:
### Summary
|             |   |
|-------------|---|
| Description | Deploy proxies transactions have same nonce  |
| Jira issue  | https://energyweb.atlassian.net/browse/MYEN-310 |

#### Type of request

- [x] Bug fix

### List of Fixes
- Proxies are created in `for await` loop


packages/proxyIdentity/src/ProxyManager.ts
```diff
@@ -2,25 +2,34 @@
...
   constructor(private erc1056: string, private erc1155: string, private owner: Signer) {
     this.proxyFactory = new ContractFactory(proxyAbi, proxyBytecode, this.owner);
+    this.provider = owner.provider;
   }
 
   async createProxy(serial: string): Promise<Contract> {
+    const address = await this.proxyFactory.signer.getAddress();
     return this.proxyFactory.deploy(
       this.erc1056, this.erc1155, serial, await this.owner.getAddress(),
+      { nonce: await this.provider.getTransactionCount(address) },
```
ni3gavhane, line 23: I hope this is solving this problem. Otherwise blockHash of the recent block could be an option.

ni3gavhane (approved):

packages/proxyIdentity/src/ProxyManager.ts
JGiter: yes. this seems to be working. I consoled tx nonces and they are indeed different

PR #252, opened 2020-11-16
feature/resolver fetch events from specific block

JGiter:
### Summary
|             |   |
|-------------|---|
| Description | Provide API to retrieve events starting from specified block |
| Jira issue  | https://energyweb.atlassian.net/browse/MYEN-349 |

#### Type of request

### List of Features
- added `Resolver.readFromBlock(did, block)` method
- `Operator.update` will return block number to use in conjunction with `Resolver.readFromBlock`
- `Operator.deactivate` will throw instead of returning false in case of failure

### DIFF
--- a/packages/did-document/src/full/documentFull.ts
+++ b/packages/did-document/src/full/documentFull.ts
@@ -1,6 +1,7 @@
 import {
   DIDAttribute, IOperator, IUpdateData, PubKeyType,
 } from '@ew-did-registry/did-resolver-interface';
+import { utils } from 'ethers';
 import { IDIDDocumentFull } from './interface';
 import { DIDDocumentLite } from '../lite';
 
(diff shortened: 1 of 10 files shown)


packages/did-document/src/full/documentFull.ts
```diff
@@ -42,8 +43,8 @@ class DIDDocumentFull extends DIDDocumentLite implements IDIDDocumentFull {
    * ```
    * @return { boolean }
    */
-  async deactivate(): Promise<boolean> {
-    return this._operator.deactivate(this.did);
```
ni3gavhane, line 46: how come this was working previously? async?

JGiter: return value from async function doesn't need to be prepended with `await`

ni3gavhane (approved):

PR #257, opened 2020-11-25
test and initial fix for undefined serviceEndpoint ids

jrhender:
### Summary
|             |   |
|-------------|---|
| Description | Inclusion of the "id" property of IAttributePayload is not enforced, however resolution of the Service Endpoint  expects that "id" is present. If no "id" is given, a key of "undefined" is given to the service endpoint in `handleAttributeChange`. If several service endpoints without ids are in the events, then only the most recent will be in the resolved document.  |
| Jira issue  | none  |

#### Type of request

### List of Features
This PR adds:
- A test in `did-document-full.test.ts` which demonstrates the issue
- A possible fix that uses the "serviceEndpoint" as a default "id" if none is present. Probably a better solution (maybe in addition to this change), would be to enforce that "id" is provided when creating a service endpoint.

### DIFF
--- a/packages/did-ethr-resolver/src/implementations/operator.ts
+++ b/packages/did-ethr-resolver/src/implementations/operator.ts
@@ -3,6 +3,7 @@
 import {
   Contract, ethers, Event, utils,
 } from 'ethers';
+import { v4 as uuid } from 'uuid';
 import {
   Algorithms,
   DIDAttribute,
@@ -92,7 +93,7 @@ export class Operator extends Resolver implements IOperator {
       algo: Algorithms.Secp256k1,
       type: PubKeyType.VerificationKey2018,
       encoding: Encoding.HEX,
-      value: { publicKey: `0x${this.getPublicKey()}`, tag: KeyTags.OWNER },
+      value: { id: uuid(), publicKey: `0x${this.getPublicKey()}`, tag: KeyTags.OWNER },
     };
     const validity = 10 * 60 * 1000;
     await this.update(did, attribute, updateData, validity);
@@ -350,6 +351,7 @@ export class Operator extends Resolver implements IOperator {
       const publicKeyTag = pk.id.split('#')[1];
 
       const keyValue: IAttributePayload = {
+        id: 'NEED TO DETERMINE',
         publicKey: value,
         tag: publicKeyTag,
       };
--- a/package-lock.json
+++ b/package-lock.json
@@ -127,9 +127,9 @@
       }
     },
     "@babel/helper-create-class-features-plugin": {
-      "version": "7.13.10",
-      "resolved": "https://registry.npmjs.org/@babel/helper-create-class-features-plugin/-/helper-create-class-features-plugin-7.13.10.tgz",
-      "integrity": "sha512-YV7r2YxdTUaw84EwNkyrRke/TJHR/UXGiyvACRqvdVJ2/syV2rQuJNnaRLSuYiop8cMRXOgseTGoJCWX0q2fFg==",
+      "version": "7.13.11",
+      "resolved": "https://registry.npmjs.org/@babel/helper-create-class-features-plugin/-/helper-create-class-features-plugin-7.13.11.tgz",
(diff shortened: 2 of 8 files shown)


jrhender:
I noticed in the documentation comment for `IAttributePayload` that it says we should move towards using `IPublicKey` or `IServiceEndpoint`. Would `IAuthentication` be included in that as well?

JGiter (commented):
This solution will work if service endpoints uniquely identified by `serviceEndpoint` field. But if for example one service is used for several purposes (that is there are several services of different types but with the same `serviceEndpoint`) then only one of them will be in the read document. So I suggest just to make `id` required in `IAttributePayload` instead of using natural keys or their composition.

JGiter:
Regarding moving from `IAttributePayload`. If we will replace this then we probably will need `IAthentication` as well, but since a lot of code depends on current API I don't think such change will happen

jrhender:
@ni3gavhane @JGiter , I looked at the Decentralized Identity Foundation's ethr-did-resolver implementation, it looks like they are constructing the service/publickey "keys" (i.e. ids) from the event data:
```const key = `${event._eventName}-${event.delegateType || event.name}-${event.delegate || event.value}```
https://github.com/decentralized-identity/ethr-did-resolver/blob/ecfd8e8568b906d2fab35ab4114b4e29565faa96/src/ethr-did-resolver.js#L73

What do you think of their approach? If we were to do something similar, I guess we would remove `id` from `IAttributePayload` and then the id/key would be determined from the event data. I'm not sure if this would still have the problem of duplicate keys. I'm not sure if the event data for two identical publickey/services `IUpdateData` would generate the same key using above code snippet 🤔 . If so, maybe we could add the transaction nonce as well (or something similar) to ensure uniqueness?
I guess it also depends on whether or not it is a requirement that users be able to set their own id/key?

**Comment by ni3gavhane** [comment]
> @ni3gavhane @JGiter , I looked at the Decentralized Identity Foundation's ethr-did-resolver implementation, it looks like they are constructing the service/publickey "keys" (i.e. ids) from the event data:
> `` const key = `${event._eventName}-${event.delegateType || event.name}-${event.delegate || event.value} ``
> https://github.com/decentralized-identity/ethr-did-resolver/blob/ecfd8e8568b906d2fab35ab4114b4e29565faa96/src/ethr-did-resolver.js#L73
> 
> What do you think of their approach? If we were to do something similar, I guess we would remove `id` from `IAttributePayload` and then the id/key would be determined from the event data. I'm not sure if this would still have the problem of duplicate keys. I'm not sure if the event data for two identical publickey/services `IUpdateData` would generate the same key using above code snippet  . If so, maybe we could add the transaction nonce as well (or something similar) to ensure uniqueness?
> I guess it also depends on whether or not it is a requirement that users be able to set their own id/key?

Thanks John. I have seen this earlier. But I would still go ahead with the approach we discussed, its simple. We can leave it with the user and application to take care of the unique id for the serviceEndpoint.

File: packages/did-ethr-resolver/src/implementations/operator.ts [hunk 1]
```diff
@@ -92,7 +93,7 @@ export class Operator extends Resolver implements IOperator {
       algo: Algorithms.Secp256k1,
       type: PubKeyType.VerificationKey2018,
       encoding: Encoding.HEX,
-      value: { publicKey: `0x${this.getPublicKey()}`, tag: KeyTags.OWNER },
+      value: { id: uuid(), publicKey: `0x${this.getPublicKey()}`, tag: KeyTags.OWNER },
```
jrhender, line 96: @JGiter Do you think it makes sense to use `uuid` as the default id for publicKeys?

JGiter: actually id of public constructed from `did` and public key tag: https://github.com/energywebfoundation/ew-did-registry/blob/b3e025141fb085b1a468a94de775178014fa8c86/packages/did-ethr-resolver/src/functions/functions.ts#L109 Why not use same approach for service endpoint id? For example like `${did}#${serviceEndpoint}`

jrhender:
Closing PR as it is stale

PR #269, opened 2021-03-04
feature/offerable identity

JGiter:
| Jira issue  |  https://energyweb.atlassian.net/browse/MYEN-694  |

`ProxyIdentity` redesigned as `OfferableIdentity`, which allows to offer identity with possibility to accept or reject it. For use cases check tests in the `proxyidentity` package

Design decisions:
 - Identity owner doesn't control identity document directly. Identity owns its document
 - Identity owner can offer identity by `OfferableIdentity.offer`
 - Owned identity can be operated by `OfferableIdentityOperator` which extends `Operator` and therefore can be used in claims workflow. Some `Operator`s methods are excluded:
    -  `changeOwner` - to prevent identity transfer without acknowledgment
    -  `revokeDelegate` - notion of delegate is not defined on offerable identities


jrhender:
Looks good to me so far 😃

packages/proxyIdentity/contracts/offerableIdentity.sol
```diff
@@ -0,0 +1,65 @@
...
+  function reject() external isOfferedTo {
+    IdentityManager(manager).identityRejected(address(this), offeredTo);
+    closeOffer();
+  }
+
+  function closeOffer() internal {
+    offeredTo = address(0);
+  }
+
+  function sendTransaction(bytes memory _data, address to, uint256 value)
```
jrhender, line 53: Does this function need the `isOwner` modifier?

packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -0,0 +1,45 @@
+pragma solidity ^0.5.0;
+
+
+contract IdentityManager {
```
jrhender, line 4: If this contract is just doing the events, maybe it's more specific to call it `IdentityEvents` or maybe `IdentityNotifier`

packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -0,0 +1,45 @@
+pragma solidity ^0.5.0;
```
jrhender, line 1: Why use the `0.5` compiler instead of a higher version (for both this contract and `OfferableIdentity`)?
I'm not saying it should be higher, I'm not too sure. But it maybe a higher version would mean a contract inheriting this contract could use new features?

jrhender (commented):
Looks good, just a few questions

packages/proxyIdentity/contracts/offerableIdentity.sol
JGiter: If tranasaction sent to ERC1056, then ownership will be checked by ERC1056, but someone who is not the owner might use this method to send tx from asset to some arbitrary contract what potentially could be dangerous. That is why I added this additional safe guarantee.

packages/proxyIdentity/contracts/IdentityManager.sol
JGiter: I plan to expand this contract in the future, such as adding methods to return a list of assets or may be batch asset creation

packages/proxyIdentity/contracts/IdentityManager.sol
JGiter: Good idea. I'll upgrade version

packages/proxyIdentity/contracts/offerableIdentity.sol
jrhender: Please add `isOwner` modifier

jrhender (changes requested):

jrhender (approved):
Looks good

jrhender:
@JGiter I was taking a look at ERC725X and wondering if maybe `OfferableIdentity` should be able to create contracts as well. https://github.com/ERC725Alliance/ERC725/blob/master/implementations/contracts/ERC725/ERC725X.sol Maybe it's overkill but just thought I would throw it out there. What do you think?

JGiter:
> @JGiter I was taking a look at ERC725X and wondering if maybe `OfferableIdentity` should be able to create contracts as well. https://github.com/ERC725Alliance/ERC725/blob/master/implementations/contracts/ERC725/ERC725X.sol Maybe it's overkill but just thought I would throw it out there. What do you think?

May be active assets should be able to do this. If so then it make sense to add such feature in separate ActiveIdentity contract, which extends OfferableIdentity

PR #271, opened 2021-03-15
bug/not compiled in strict mode

JGiter:
| Jira issue  |  https://energyweb.atlassian.net/browse/MYEN-730  |

Fixes typescript definition issues, which doesn't allow compile dependent projects in strict mode:
- dependencie's types moved from devDeps to Deps
- `sjcl-complete` aliased as `sjcl` to allow use `sjcl`'-s type definitions
- library compiled in strict mode

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,12 +1,13 @@
 {
   "compilerOptions": {
+    "strict": true,
     "module": "commonjs",
     "target": "es5",
     "declaration": true,
     "sourceMap": true,
     "noImplicitAny": true,
     "resolveJsonModule": true,
-    "skipLibCheck": true,
+    "skipLibCheck": false,
     "esModuleInterop": true,
     "lib": [
       "ESNext",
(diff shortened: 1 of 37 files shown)


jrhender (changes requested):
If the errors have been resolved, I think that `strict` should be turned on in `tsconfig.json` so that future issues aren't introduced

tsconfig.json
```diff
@@ -1,12 +1,13 @@
 {
   "compilerOptions": {
+    "strict": true,
```
jrhender, line 3: 🙏  👍

tsconfig.json
```diff
@@ -1,12 +1,13 @@
...
     "module": "commonjs",
     "target": "es5",
     "declaration": true,
     "sourceMap": true,
     "noImplicitAny": true,
     "resolveJsonModule": true,
-    "skipLibCheck": true,
+    "skipLibCheck": false,
```
jrhender, line 10: 👍

jrhender (approved):

PR #276, opened 2021-03-24
feature/offerable identity

JGiter:
| [Jira issue](https://energyweb.atlassian.net/browse/MYEN-690) | [Specification](https://energyweb.atlassian.net/wiki/spaces/UAAR/pages/971538591/RFC+197+Proxy+smart+contract+specification) |


packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -1,12 +1,40 @@
-pragma solidity 0.8.0;
+pragma solidity 0.4.26;
 
+import "@optionality.io/clone-factory/contracts/CloneFactory.sol";
+import "openzeppelin-solidity/contracts/introspection/ERC165Checker.sol";
+import "./OfferableIdentity.sol";
+
+
+contract IdentityManager is CloneFactory {
+  address libraryAddress;
```
jrhender, line 9: This is the address of the implementation of `OfferableIdentity` that is being cloned, right? Maybe a name like `IdentityImplementation` might be a little more specific?

packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -1,12 +1,40 @@
-pragma solidity 0.8.0;
+pragma solidity 0.4.26;
 
+import "@optionality.io/clone-factory/contracts/CloneFactory.sol";
+import "openzeppelin-solidity/contracts/introspection/ERC165Checker.sol";
```
jrhender, line 4: Looks like you're using this `CloneFactory` from optionality.io. I'm not two familiar with it. Why use it instead of the Clones library provided by OpenZeppelin? https://docs.openzeppelin.com/contracts/4.x/api/proxy#Clones

packages/proxyIdentity/contracts/OfferableIdentity.sol
```diff
@@ -1,19 +1,33 @@
-pragma solidity 0.8.0;
+pragma solidity 0.4.26;
 
+import "./IERC165.sol";
 import "./IdentityManager.sol";
 
 
-contract OfferableIdentity {
+contract OfferableIdentity is IERC165 {
```
jrhender, line 7: Why using our own `IERCI65`  interface instead of the OpenZeppelin `IERC165` interface or (perhaps even better) inheriting from the `ERC165` contract? https://docs.openzeppelin.com/contracts/4.x/utilities#api:introspection.adoc#ERC165

scripts/deployIdentityMlanaget.js
```diff
@@ -0,0 +1,14 @@
+const { JsonRpcProvider } = require('ethers/providers');
+
+const { ContractFactory, Wallet } = require('ethers');
+const { abi, bytecode } = require('../packages/proxyIdentity/build/contracts/IdentityManager.json');
+
+const deployer = new Wallet(
+  '9d69bf3c7bda0fc771a59818e5a7f894f3b4fbd3114891db3d4de7f46aa71ca8',
+  new JsonRpcProvider('https://volta-rpc-vkn5r5zx4ke71f9hcu0c.energyweb.org/'),
```
jrhender, line 8: IMO, we should just use volta-rpc.energyweb.org here. I know we are using it publicly in Switchboard, but I think it is supposed to be an internal endpoint so better not to have it checked into ew-did-registry code

packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -1,12 +1,40 @@
-pragma solidity 0.8.0;
+pragma solidity 0.4.26;
```
jrhender, line 1: Is the solidity downgrade so that we can use the optionality CloneFactory?

packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -16,38 +44,70 @@ contract IdentityManager {
...
+  }
+  
+  function identityCreated(address _identity, address _owner)
     external
-    isIdentity(msg.sender, identity)
+    isIdentity(msg.sender, _identity) 
+    isOfferable
   {
-    emit IdentityCreated(identity, owner, block.timestamp);
+    if (created[msg.sender]) {
```
jrhender, line 77: 👍  Nice, I think it is smart to have created be separate from verified as you have done

jrhender (commented):
Mainly just some questions about the use of OpenZeppelin ERC165 and Clones and optionality.io CloneFactory

packages/proxyIdentity/build/contracts/IERC165.json
```diff
@@ -0,0 +1,306 @@
...
+  ],
+  "metadata": "",
+  "bytecode": "0x",
+  "deployedBytecode": "0x",
+  "sourceMap": "",
+  "deployedSourceMap": "",
+  "source": "pragma solidity 0.4.26;\n\n\ninterface IERC165 {\n  function supportsInterface(bytes4 interfaceID) external pure returns (bool);\n}",
+  "sourcePath": "/home/dima/Development/ew-did-registry/packages/proxyIdentity/contracts/IERC165.sol",
+  "ast": {
+    "absolutePath": "/home/dima/Development/ew-did-registry/packages/proxyIdentity/contracts/IERC165.sol",
```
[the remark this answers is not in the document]
JGiter: I think that they are not important because only `abi` and `bytecode` are exported

packages/proxyIdentity/contracts/IdentityManager.sol
JGiter: Yes, optionality restricts soliditiy by v0.4

packages/proxyIdentity/contracts/IdentityManager.sol
JGiter: Didn't know they added it. After looking at discussion https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1162 I decided they are not going to add it in near future. I will try to use it

packages/proxyIdentity/contracts/IdentityManager.sol
JGiter: It is address of contract containing logic of all clones. Clones (which has only state) will delegate all calls to it. So it is used as library

packages/proxyIdentity/contracts/OfferableIdentity.sol
```diff
@@ -0,0 +1,101 @@
...
+  function supportsInterface(bytes4 interfaceID)
+    public
+    virtual
+    view
+    override
+    returns (bool)
+  {
+    return interfaceID == ERC165ID || interfaceID == offerableID;
+  }
+}
```
jrhender, line 101: The openzeppelin `ERC165` contract provides `supportsInterface` function. Do we need this function as well?

jrhender (commented):
Just one question about the usage of the ERC165 contract

jrhender:
@JGiter I think the contracts are looking good. Could you also request a review from @ngyam please?

ngyam:
coolio, giving it a look tomo

packages/proxyIdentity/contracts/OfferableIdentity.sol
JGiter: Right. We can reuse super to check support for ERC165

packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -1,95 +1,114 @@
...
+import "./IOfferable.sol";
+
 
 contract IdentityManager {
-  mapping(address => address) owners;
+  address libraryAddress;
+  address ethrRegistry;
+  address owner;
+  
+  mapping(address => bool) created;
```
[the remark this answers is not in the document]
ngyam: ABI coder v2 is activated by default since 0.8 so it's possible to have a getter for a struct too

ngyam (changes requested):
I have no knowledge at all about the context or the requirements so I don't know what is going on and why but I tried to give it a quick read. See my comments below.

General ones:
- if we use 0.8, let's use 0.8.3 the latest version
- build folder is usually not pushed
- please use a linter cause they can also suggest some minor things
 
I asked Micha for a MythX subscription so soon we will have deep scanning to catch some vunerabilities automatically.

Since all this stuff will be audited, I would ask you guys to produce a proper spec document, that explains all to a person who doesn't know anything about the project. What is the goal, why, what were the design choices and why, what are the components etc. Th auditors will ask for it, so it's a must have for both internal/external auditing

jrhender:
> Since all this stuff will be audited, I would ask you guys to produce a proper spec document, that explains all to a person who doesn't know anything about the project. What is the goal, why, what were the design choices and why, what are the components etc. Th auditors will ask for it, so it's a must have for both internal/external auditing

Thanks for the review @ngyam ! So many good points. Do you have any examples of spec documents? Do you mean like the ERC/EIP documents?

ngyam:
> > Since all this stuff will be audited, I would ask you guys to produce a proper spec document, that explains all to a person who doesn't know anything about the project. What is the goal, why, what were the design choices and why, what are the components etc. Th auditors will ask for it, so it's a must have for both internal/external auditing
> 
> Thanks for the review @ngyam ! So many good points. Do you have any examples of spec documents? Do you mean like the ERC/EIP documents?

yes, sending you the link in slack

packages/proxyIdentity/contracts/OfferableIdentity.sol
```diff
@@ -0,0 +1,92 @@
+pragma solidity 0.8.0;
+
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import "./IdentityManager.sol";
+import "./IOfferable.sol";
+
+
+contract OfferableIdentity is IOfferable, ERC165 {
+  address public owner;
```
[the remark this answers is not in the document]
ngyam: if owner never changes and set in constructor -> mark immutable

packages/proxyIdentity/contracts/IdentityManager.sol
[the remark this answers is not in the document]
JGiter: I would like to keep `created` and `offered` flags private

packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -1,95 +1,114 @@
...
+import "@openzeppelin/contracts/utils/introspection/ERC165Checker.sol";
+import "./OfferableIdentity.sol";
+import "./IOfferable.sol";
+
 
 contract IdentityManager {
-  mapping(address => address) owners;
+  address libraryAddress;
+  address ethrRegistry;
+  address owner;
```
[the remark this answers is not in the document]
JGiter: I don't think we need it because it mostly adds logic intended to change owner that is not planned for Identity Manager

packages/proxyIdentity/contracts/OfferableIdentity.sol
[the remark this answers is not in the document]
JGiter: It can be changed

packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -1,95 +1,114 @@
...
-    address indexed offeredTo,
-    uint256 indexed at
-  );
-  event IdentityOfferCanceled(
-    address indexed identity,
-    address indexed owner,
-    address oferedto,
-    uint256 indexed at
-  );
+  event IdentityCreated(address indexed identity, address indexed owner, uint256 indexed at);
```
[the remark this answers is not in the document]
JGiter: We need to index `at` to filter events

JGiter:
Much appreciated for help @ngyam!

JGiter:
@ngyam I've updated RFC with specs about project - https://energyweb.atlassian.net/wiki/spaces/UAAR/pages/971538591/RFC+197+Proxy+smart+contract+specification#Functional-requirements

.eslintrc.json
```diff
@@ -46,7 +46,9 @@
     "no-unused-expressions": 0,
     "import/no-extraneous-dependencies": ["error", {"devDependencies": true}],
     "chai-friendly/no-unused-expressions": 2,
-    "func-names": "off"
+    "func-names": "off",
+    "no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": ["error"]
```
jrhender, line 51: Does this mean that an error is still emitted if there are unused vars? Was this so that you could disable eslint in `didStore.delete()` in this PR?

packages/proxyIdentity/package.json
```diff
@@ -24,13 +27,14 @@
     "ethers": "4.0.45"
   },
   "devDependencies": {
-    "@openzeppelin/contracts": "^2.5.0",
+    "@openzeppelin/contracts": "4.0.0",
```
jrhender, line 30: 👍

.eslintrc.json
JGiter: Its not exactly related to this PR, I just noticed that eslint only warns about unused variables and allows to commit which could lead to troubles. So I decided to be more explicit when such variables are allowed

jrhender (approved):

packages/proxyIdentity/contracts/OfferableIdentity.sol
```diff
@@ -0,0 +1,89 @@
...
+  function update(bytes memory data, uint256 value)
+    external
+    isOwner
+    override
+    returns (bool success)
+  {
+    (success, ) = ethrRegistry.call{ gas: gasleft() - 5000, value: value}(data);
+    require(success, "OfferableIdentity: Document is not updated");
+    emit TransactionSent(data, value);
+  }
```
jrhender, line 76: @JGiter After speaking with Micha, he confirmed that this should be a generic proxy function (as you had it before). Could you please revert it to the generic function? This is for a couple of reasons:

- The asset should be able to accommodate future use cases (such as being able to notify of energy production)
- If the `ethrRegistry` needs to be redeployed then we will not need to redeploy all of the asset contracts 

@ngyam Does this seem reasonable to you?

jrhender (changes requested):

packages/proxyIdentity/contracts/OfferableIdentity.sol
JGiter: Reverted. Also added setter to avoid redeploying when registry is changed

JGiter: Reverted. To allow call arbitrary contract I just added its address to params

packages/proxyIdentity/contracts/OfferableIdentity.sol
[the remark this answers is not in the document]
ngyam: ah ok I see it's set in init

packages/proxyIdentity/contracts/IdentityManager.sol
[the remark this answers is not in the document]
ngyam: If this is something that needs to be owned by EWF in the end, then it's better to have it transferrable.

ngyam (changes requested):
Almost there, some open comments to address

packages/proxyIdentity/contracts/OfferableIdentity.sol
ngyam: If you have a good reason for this, sure. It looks okay like this.

packages/proxyIdentity/contracts/IdentityManager.sol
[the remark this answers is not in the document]
JGiter: Ok. I've grouped identity's props in Identity struct

packages/proxyIdentity/contracts/IdentityManager.sol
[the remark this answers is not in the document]
JGiter: Since ethrRegistry now will be provided by `sendTransaction` (former `update`) caller and after discussion we decided to not allow to change identity library there is no need for IdentityManager owner any longer

jrhender (approved):
Looks good to me.
 @ngyam Just a note, if hasn't been stated already, that we decided to keep the contract build JSON files in source control so that Switchboard can retrieve them by GitHub URL.

packages/proxyIdentity/contracts/IdentityManager.sol
```diff
@@ -1,95 +1,114 @@
...
-      identity,
-      owners[identity],
-      offeredTo,
-      block.timestamp
-    );
+    offered[msg.sender] = false;
+    emit IdentityOfferCanceled(msg.sender, owners[msg.sender], _offeredTo, block.timestamp);
   }
-}
+}
```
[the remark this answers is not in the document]
ngyam: pingg

ngyam (changes requested):
one little extra endline missing that triggers my OCD, otherwise LGTM :)

PR #303, opened 2021-04-05
bug/recreated_document

JGiter:
Increases owner public key validity to maximum to avoid often document recreations


jrhender (approved):
Looks good to me. @JGiter What do you think the previous rational was on setting a limited validity period? Do you think it was because it is just a general good practice to "enforce" key rotation?

JGiter:
> Looks good to me. @JGiter What do you think the previous rational was on setting a limited validity period? Do you think it was because it is just a general good practice to "enforce" key rotation?

Yes, at the time when this was added, it was not known how significantly updates to the document would increase the duration of its reading. By the way now validity is also limited, just much bigger. This value required by Ethereum registry contract

PR #306, opened 2021-04-26
feature/export erc1056 validation

JGiter:


packages/did/package.json
```diff
@@ -23,5 +23,8 @@
     "name": "EnergyWeb",
     "url": "https://www.energyweb.org/"
   },
-  "license": "GPL-3.0-or-later"
+  "license": "GPL-3.0-or-later",
+  "dependencies": {
```
jrhender, line 27: Should this be in the dev dependencies as it is being used only in tests?

JGiter: Right. Fixed

jrhender (approved):
Looks good

PR #309, opened 2021-05-13
feature/issue claim from unsigned claim data

JGiter:
### Summary
|             |   |
|-------------|---|
| Description | As there are no more requirements on claim requester, claim can be issued from row claim data without signing it. 
| Jira issue  |  |

### DIFF
--- a/packages/claims/src/claimsIssuer/claimsIssuer.ts
+++ b/packages/claims/src/claimsIssuer/claimsIssuer.ts
@@ -22,7 +22,8 @@ export class ClaimsIssuer extends Claims implements IClaimsIssuer {
    * claims = new ClaimsIssuer(issuer);
    * const issuedToken = await claims.issuePublicClaim(token);
    * ```
-   * @params { string } token to verify
+   * @param { string | IPublicClaim } claim - claim to issue. Can be
+   * specified as signed or unsinged claim
    * @returns { Promise<string> } issued token
    */
   async issuePublicClaim(claim: string | IPublicClaim): Promise<string> {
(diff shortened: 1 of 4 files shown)


jrhender (commented):
I'm not understanding the PR based on the file changes. It looks like the change is to only accept a jwt `token` instead of possibly `IPublicClaim`? But I'm not sure what that has to do with
>claim can be issued from row claim data without signing it.

JGiter:
Purpose of this change is to free requester from signing claim before submitting to issuer. I've updated doc comment

packages/claims/src/claimsIssuer/claimsIssuer.ts
```diff
@@ -22,11 +22,14 @@ export class ClaimsIssuer extends Claims implements IClaimsIssuer {
    * claims = new ClaimsIssuer(issuer);
    * const issuedToken = await claims.issuePublicClaim(token);
    * ```
-   * @params { string } token to verify
+   * @param { string | IPublicClaim } claim - claim to issue. Can be
+   * specified as signed or unsinged claim
```
jrhender, line 26: Minor typo
```suggestion
   * specified as signed or unsigned claim
```

packages/claims/src/claimsIssuer/claimsIssuer.ts
```diff
@@ -22,11 +22,14 @@ export class ClaimsIssuer extends Claims implements IClaimsIssuer {
...
    * @returns { Promise<string> } issued token
    */
-  async issuePublicClaim(token: string): Promise<string> {
-    const claim: IPublicClaim = this.jwt.decode(token) as IPublicClaim;
+  async issuePublicClaim(claim: string | IPublicClaim): Promise<string> {
```
jrhender, line 29: Could you please add a test which issues a claim using `IPublicClaim`?

jrhender (changes requested):
Makes sense now, thanks. Could you please add a test which covers this scenario?

jrhender (approved):
Looks good to me

PR #313, opened 2021-06-29
bug/SWTCH-1089 can't enroll onwed asset

JGiter:
### Summary
|             |   |
|-------------|---|
| Description | `OfferableIdentity` must signify itself as being owned in order enroll itself by owner|
| Jira issue  | https://energyweb.atlassian.net/browse/SWTCH-1089 |

### List of Features
- OfferableIdentity implements `IOwned`

Should be merged after https://github.com/energywebfoundation/iam-contracts/pull/28


packages/proxyIdentity/contracts/IOwned.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity 0.8.3;
```
jrhender, line 1: I think it would make sense to have a more flexible pragma here to allow the interface to be used by more consumers. I picked `0.6.0` as the lower bound but maybe we can go even lower, I think we should go as low as possible
```suggestion
pragma solidity >0.6.0 <0.9.0;
```

jrhender (approved):
Looks good

PR #315, opened 2021-07-19
feat(eslint): setting eslint

knzeng-e:
### Summary
|             |   |
|-------------|---|
| Description |  Add "no extraneous dependencies" es-lint plugin|
| Jira issue  | https://energyweb.atlassian.net/browse/EDR-7  |

#### Pull Request targuet

- [x] activating no-extraneous-dependencies rule
- [x] tackling eslint warnings/errors
- [x] running eslint check on travis

### List of Features
- Enabling eslint plugin benmosher/eslint-plugin-import  with this rule https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md 
- Enable linting on CI level


jrhender (approved):
Looks good, thanks Kevin 👍

jrhender (commented):
Why the removal of the root `package-lock.json`?

jrhender (changes requested):
Just want to clarify the package-lock.json rational before merging

knzeng-e:
> Why the removal of the root `package-lock.json`?

That was only to fire a new commit to see if the `deploy` step would trigger. But maybe there's a cleaner way to make a commit without changing anything ?

jrhender:
> > Why the removal of the root `package-lock.json`?
> 
> That was only to fire a new commit to see if the `deploy` step would trigger (it is blocking and I wasn't able to merge after your aproval). But maybe there's a cleaner way to make a commit without changing anything ?

I think you need to change a file to make a commit

jrhender (approved):

PR #319, opened 2021-08-16
Fix/edr 15/update doc offerable identity

jrhender:
### Summary

|             |   |
|-------------|---|
| Description | Core of change is that getting proxyIdentity package tests to run during CI and pass with ethers v5   |
| Jira issue  | EDR-14, EDR-15   |

#### Type of request

### List of Features

- [x] Run offerable-identity tests as part of "npm run test"
- [x] Add root package-lock.json
- [x] Fix params issue in offerableIdentityOperator
- [x] `validity` parameter is not included in `revokeAttribute` transaction sent by `OfferableIdentityOperator`
- [x] `EthereumDIDRegistry` method to invoke will be identified by its name instead of resolving from updated DID attribute
- [x] Ensure that offerable-idenity tests are passing in CI (might need to add contract build)


knzeng-e (dismissed):
Looks good to me, Thank you @jrhender !

JGiter (commented):
I suggest use same npm script names as in other packages, which are `test` and `test-rpc`. I included this change  in proposed commits
Also these commits fixes two flaws, which I've added to PR description

jrhender:
@JGiter Why this change?
>  EthereumDIDRegistry method to invoke will be identified by its name instead of resolving from updated DID attribute

JGiter:
> @JGiter Why this change?
> 
> > EthereumDIDRegistry method to invoke will be identified by its name instead of resolving from updated DID attribute

`OfferableIdentityOperator._sendTransaction` doesn't determine `revokeAttribute` method (https://github.com/energywebfoundation/ew-did-registry/blob/bcba5f5344eddf8b05dc5f500c026f20f14c860c/packages/proxyIdentity/src/offerableIdentityOperator.ts#L89). It can be detected by comparing validity with 0 or `revokeAttribute` can be overridden, but for me the most straightforward way is explicitly set method name. Initial approach of passing in `Function` to `_sendTransaction` leads to unnecessary  complications

PR #321, opened 2021-08-28
bug/EDR-30 Can't extend EwSigner from Signer

JGiter:
### Summary

| Description | Instantiation of EwSigner throws `constructor can not be invoked without new` |
| Jira issue  | https://energyweb.atlassian.net/browse/EDR-30 |

#### Type of request

- [x] Bug fix

Fixes:
- compile to es6 to enable extending from es6 classes such as `ethers.Signer`
- As repository should contain only source code contract artifacts from proxy identity will be build when installing package

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ node_modules
 **/*.map
 docs
 build
-!packages/proxyIdentity/build
 dist
 *.d.ts
 !typings/**/*.d.ts
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "strict": true,
     "module": "commonjs",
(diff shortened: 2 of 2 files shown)


knzeng-e (approved):

.gitignore
```diff
@@ -3,7 +3,6 @@ node_modules
 **/*.map
 docs
 build
-!packages/proxyIdentity/build
```
jrhender, line 6: Why including `packages/proxyIdentity/build` now?

jrhender (approved):

.gitignore
JGiter: I propose following approach chosen for iam-contracts, namely building artifacts when installing package rather then storing them in repository

PR #326, opened 2021-09-08
Feat/edr 37/ew json rpc signer from provider

jrhender:
### Summary

|             |   |
|-------------|---|
| Description | Collapses IdentityOwner into EwSigner. EwSigner was only used for IdentityOwner so no need to have extra layer of abstraction. This means that public key is now required for creation of EwSigner. Given that public key is required, no need for extensions of EwSigner.  |
| Jira issue  | https://energyweb.atlassian.net/browse/EDR-37  |


packages/did-ethr-resolver/src/implementations/ewJsonRcpSigner.ts
```diff
@@ -1,23 +1,32 @@
...
+  /**
+   * Allows the creation of EwJsonRpcSigner using an ethers library Provider.
+   * This is convenient if a suitable ethers provider is available.
+   * If instead an EIP1993 provider is available, see {@linkcode fromEIP1193}
+   * @param provider an ethers JsonPrcProvider or Web3Provider, with an associated signer
+   */
+  // eslint-disable-next-line max-len
+  public static fromEthersProvider(provider: providers.JsonRpcProvider | providers.Web3Provider): EwJsonRpcSigner {
+    return new EwJsonRpcSigner(provider);
+  }
```
jrhender, line 28: This factory method is a little redundant because it simply delegates to the constructor, but it is maybe less confusing for consumers of the API to know to look for factory methods for creation instead of mixing constructor and factory method

jrhender: What do you think @JGiter @knzeng-e?

JGiter: Let it be. Constructor might be changed later on. Also I don't really understand purpose of `EwJsonRpcSigner` and `IdentityOwner`. Could you please elaborate what prevents from defining `fromPrivateKey`, `fromEthersProvider` and `fromEIP1193Provider` directly on `EwSigner`?
And what is the reason of `JsonRpcProvider | Web3Provider` type?

jrhender: - I think we could put `fromEthersProvider` and `fromEIP1193Provider` directly on `EwSigner`. However, `EwPrivateKeySigner` has additional `public` and `private` key properties. If that is not useful to have then we can also put `fromPrivateKey` on `EwSigner`.
- `IdentityOwner` extends `EwSigner` with a `publicKey` property. That is its purpose from my perspective.
- Ah, you're right, I see that `Web3Provider` extends `JsonRpcProvider`, so we can remove `Web3Provider` from the type: https://github.com/ethers-io/ethers.js/blob/f2a32d0d5b4ea3721d3f3ee14db56e0519cf337f/packages/providers/src.ts/web3-provider.ts#L136

packages/did-ethr-resolver/src/implementations/ewSigner.ts
```diff
@@ -1,12 +1,24 @@
...
 /**
- * A base signer class that encapsulates the ethers Signer.
- * The purpose of this encapsulation is to allow consumers more flexiblity in ethers version.
+ * A signer class that encapsulates the ethers Signer and ensures that publicKey is available.
+ * The public is necessary for verification of signatures made by the signer.
+ * The purpose of the ethers encapsulation is to allow consumers more flexiblity in ethers version.
  */
-export abstract class EwSigner extends Signer {
-  constructor(
+export class EwSigner extends Signer {
```
jrhender, line 14: Core of change is here

JGiter (dismissed):

PR #328, opened 2021-09-17
feat(EwSigner): add fromEthersSigner

jrhender:
### Summary

add `fromEthersSigner` also removed fromEthersProvider as it is not necessary to have. It makes more sense to create from signer as we are creating a EW"Signer". 
Also, added check that signer is actually available if creating from EIP1193. This is kinda of the issue that the EKC team had where they had a provider that wasn't actually connected to a signer. So better to check early.
|             |   |
|-------------|---|
| Jira issue  | https://energyweb.atlassian.net/browse/EDR-43 |

#### Type of request

- [x] New feature  **It is a breaking change** because removed `fromEthersProvider`

### DIFF
--- a/packages/did-ethr-resolver/src/implementations/ewSigner.ts
+++ b/packages/did-ethr-resolver/src/implementations/ewSigner.ts
@@ -76,15 +76,14 @@ export class EwSigner extends Signer {
   }
 
   /**
-   * A factory method to create an EwSigner using an ethers library Provider.
-   * This is convenient if a suitable ethers provider is available.
+   * A factory method to create an EwSigner using an ethers library Signer.
+   * This is convenient if a suitable ethers signer is available.
    * If instead an EIP1993 provider is available, see {@linkcode fromEIP1193}
-   * @param provider an ethers JsonPrcProvider or Web3Provider, with an associated signer
+   * @param signer an ethers JsonPrcProvider or Web3Provider, with an associated signer
    * @param publicKey the publicKey of the signer associated with the provider
    */
-  // eslint-disable-next-line max-len
-  public static fromEthersProvider(provider: providers.JsonRpcProvider, publicKey: string): EwSigner {
-    return new EwSigner(provider.getSigner(), publicKey);
+  public static fromEthersSigner(signer: Signer, publicKey: string): EwSigner {
+    return new EwSigner(signer, publicKey);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


yartman (commented):
👍

packages/did-ethr-resolver/src/implementations/ewSigner.ts
```diff
@@ -76,15 +76,14 @@ export class EwSigner extends Signer {
   }
 
   /**
-   * A factory method to create an EwSigner using an ethers library Provider.
-   * This is convenient if a suitable ethers provider is available.
+   * A factory method to create an EwSigner using an ethers library Signer.
+   * This is convenient if a suitable ethers signer is available.
    * If instead an EIP1993 provider is available, see {@linkcode fromEIP1193}
-   * @param provider an ethers JsonPrcProvider or Web3Provider, with an associated signer
+   * @param signer an ethers JsonPrcProvider or Web3Provider, with an associated signer
```
JGiter, line 82:
```suggestion
   * @param signer an ethers Signer connected to chain
```

packages/did-ethr-resolver/src/implementations/ewSigner.ts
```diff
@@ -104,8 +103,19 @@ export class EwSigner extends Signer {
...
+  // eslint-disable-next-line max-len
+  public static async fromEIP1193(eip1993Provider: EIP1193ProviderType, publicKey: string): Promise<EwSigner> {
     const provider = new providers.Web3Provider(eip1993Provider);
-    return new EwSigner(provider.getSigner(), publicKey);
+    const signer = provider.getSigner();
+    try {
+      // Call getAddress() to check that signer has address
+      // Signer could be returned by not actually have an address
+      // Best to check now and "fail fast"
+      await signer.getAddress();
```
JGiter, line 114: Signer can have address event not being connected, for example Wallet. Better check provider set or chain id

JGiter:
```suggestion
      await signer.getChainId();
```

JGiter: Also it is better to put this check in `EwSigner` constructor to not duplicate it in all factories

JGiter (changes requested):

packages/did-ethr-resolver/src/implementations/ewSigner.ts
jrhender: - In this case, I think we know that the `signer` has a `provider` because the `signer` is coming from `provider.getSigner()`
- I don't think it can be put in the constructor as it is an `async` call?

packages/did-ethr-resolver/src/implementations/ewSigner.ts
jrhender: Thanks 👍

packages/did-ethr-resolver/src/implementations/ewSigner.ts
JGiter: Right. Initially I misunderstood purpose of this. Its ok then. Though I think it is not responsibility of the did-reg to verify correctness of the signer implementation

JGiter (approved):

PR #331, opened 2021-09-28
test(claimsUser): verification of key with hex prefix

jrhender:
### Summary

- [x] write claims verification test
- [x] write ownerPubKey test
- [x] fix issues

|             |   |
|-------------|---| 
| Description | fix(resolver): readOwnerPubKey() handle key with 0x. Keys in DID document can have 0x prefix and so need to handle this case. Note that new didDocs are created with a key without the prefix   |
| Jira issue  | https://energyweb.atlassian.net/browse/ICL-105  |

#### Type of request

### DIFF
--- a/packages/claims/test/claimsUser.test.ts
+++ b/packages/claims/test/claimsUser.test.ts
@@ -4,11 +4,20 @@ import chaiAsPromised from 'chai-as-promised';
 import { Keys } from '@ew-did-registry/keys';
 import { Methods } from '@ew-did-registry/did';
 import {
-  Operator, EwSigner,
+  Operator, EwSigner, compressedSecp256k1KeyLength,
 } from '@ew-did-registry/did-ethr-resolver';
 import { DidStore } from '@ew-did-registry/did-ipfs-store';
 import { DIDDocumentFull } from '@ew-did-registry/did-document';
-import { ProviderSettings, ProviderTypes } from '@ew-did-registry/did-resolver-interface';
+import {
+  Algorithms,
+  DIDAttribute,
+  Encoding,
+  IUpdateData,
+  KeyTags,
+  ProviderSettings,
+  ProviderTypes,
+  PubKeyType,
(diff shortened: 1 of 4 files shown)


knzeng-e (approved):
Sounds good !

knzeng-e (commented):
I can't see any update on the verification process to validate both kinds of publicKey yet. 
[EDIT] I didn't see that it is a draft, my bad

JGiter (dismissed):

packages/claims/test/claimsUser.test.ts
```diff
@@ -66,7 +83,7 @@ describe('[CLAIMS PACKAGE/USER CLAIMS]', function () {
     await shutDownIpfsDaemon();
   });
 
-  it('createPublicClaim should create token with claim data', async () => {
+  xit('createPublicClaim should create token with claim data', async () => {
```
knzeng-e, line 86: What is the rationale behind those tests excludes please ?

packages/did-ethr-resolver/src/utils/crypto.ts
```diff
@@ -7,6 +7,8 @@ const {
   keccak256, hashMessage, arrayify, computePublicKey, recoverPublicKey,
 } = utils;
 
+export const compressedSecp256k1KeyLength = 66;
```
knzeng-e, line 10: What do you think if declare this in `constant.ts` instead ? Does that make sense ?

knzeng-e (approved):
Looks good to me @jrhender , just a few non blocking questions

nichonien (approved):

packages/did-ethr-resolver/src/utils/crypto.ts
jrhender: Thanks for the suggestion. That could make sense, will keep it in mind.

Harasz (commented):
LGTM

PR #333, opened 2021-09-29
chore(deploy.yml): use npm ci instead of npm install

jrhender:
npm ci does not modify the package-lock.json.
This ensures that exact packages are obtained.
This PR is motivated by this build failure: https://github.com/energywebfoundation/ew-did-registry/runs/3744856039?check_suite_focus=true


knzeng-e (approved):
Makes sens to me using npm ci in this case.

PR #337, opened 2021-09-30
Feature/edr 40 revocation lib

nichonien:
### Summary

- [x ] write revocation registry contract
- [ x] write library functions
- [x ] write test cases
- [x ] fix issues

|             |   |
|-------------|---|
| Description | Add libraries to support revocation for on-chain and off-chain claims through two separate contracts respectively.  The library will use the deployed contracts address to integrate revocation functionality in the existing systems. 
|
| Jira issue  |  https://energyweb.atlassian.net/browse/EDR-40 |

#### Type of request

- [x ] New feature (non-breaking change which adds functionality)


packages/revocation/src/contracts/RevocationRegistry.sol
```diff
@@ -0,0 +1,34 @@
...
+
+pragma solidity >=0.7.0 <0.9.0;
+    
+contract RevocationRegistry {
+    
+    struct RevokedClaim {
+        address revoker;
+        uint revokedTimestamp;
+    }
+    mapping (bytes32 => mapping(address => RevokedClaim)) revokedClaims;
```
jrhender, line 11: I know this might seems a bit of reverse deja-vu from our previous conversation 😅 . But I think that we don't want to store the revocations in this way. I think it would be better to have a mapping to an array of `RevokedClaims`
```suggestion
    mapping (bytes32 => RevokedClaim[]) revokedClaims;
```
In this case, the bytes32 is not the role namehash but rather a hash of the credential or some data that identifies the credential.
This is for a few reasons:
- Unlike with on-chain role claims, privacy is important. So I think it should not be possible to figure out that a DID had a role by looking at its revocations.
- We need to have an array of revocations for a particular credential. We can't have a single one because we don't know if a particular revoker is an honest revoker (and so we don't want them to fill the single slot with a dishonest revocation).
- This more generic way means that the contract can be used for non-role credential revocations

jrhender (changes requested):

packages/revocation/src/constants/constants.ts
```diff
@@ -0,0 +1,7 @@
+
+// Address of smart contracts on Volta
+export const RevocationRegOffChainCalims = '';
+export const RevocationRegsOnChainClaims = '';
+export const ENSRegistryAddr = '';
+export const ClaimManagerAddr = '';
+export const ERC1056Addr = '';
```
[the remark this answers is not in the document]
nichonien: Yes. Once both the `RevocationRegistry` contracts are deployed, these constants would store the addresses.

packages/revocation/src/implementations/revocation.ts
```diff
@@ -0,0 +1,179 @@
...
+  private _revocationRegistryOnChain: Contract;
+
+  private _revocationRegistryOffChain: Contract;
+
+  /**
+  * @param owner - Entity which controls revocation
+  */
+  constructor(
+    owner: EwSigner,
+    addressOnchain: string,
```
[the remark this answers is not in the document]
nichonien: Did you miss anything to add? the change looks the same. @jrhender

packages/revocation/src/implementations/revocation.ts
```diff
@@ -0,0 +1,179 @@
...
+  Contract, ethers, Event, utils,
+} from 'ethers';
+import { EwSigner, addressOf } from '@ew-did-registry/did-ethr-resolver';
+import { abi as RevocationOnChainAbi } from '../constants/RevocationRegistryOnChain.json';
+import { abi as RevocationOffChainAbi } from '../constants/RevocationRegistryOffChain.json';
+
+export class Revocation {
+  private _revocationRegistryOnChain: Contract;
+
+  private _revocationRegistryOffChain: Contract;
```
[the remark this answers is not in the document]
nichonien: Yeah, make sense. Will do the suggested changes.

packages/revocation/src/implementations/revocation.ts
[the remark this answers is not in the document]
jrhender: Capitalization difference

packages/revocation/src/implementations/revocationOffChain.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+  * checks the revocation details for a credential
+  * Returns the revokers and revocationTimeStamp for the revocations
+  *
+  * @param { string } credential - credential for which the status is to be checked
+  * @returns Promise<string>
+  */
+  async getRevocationDetail(credential: string): Promise<string[]> {
```
jrhender, line 53: I don't understand how the return type is just a `string[]` 🤔 . Isn't is an array of two string arrays? Like `Promise<Array<string[], string[]>>?

jrhender (commented):
Looks pretty good, just some comments

packages/revocation/src/implementations/revocationOnChain.ts
```diff
@@ -0,0 +1,127 @@
...
+  * @param { string } revoker - DID of the revoker
+  * @returns Promise<boolean>
+  */
+  async revokeListOfRoles(role: string, subjects: string [], revoker: string):
+  Promise<boolean> {
+    const roleHash = utils.namehash(role);
+    const revokerAddress = addressOf(revoker);
+    const revocationSubjects = [];
+    for (let _i = 0; _i < subjects.length; _i++) {
+      revocationSubjects[_i] = addressOf(subjects[_i]);
```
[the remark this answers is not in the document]
nichonien: My bad, I took this piece from the solidity and then forgot to change it.

packages/revocation/src/implementations/revocationOffChain.ts
nichonien: I tested the code, consoled the output as well and it worked. AFAIK Promise takes on argument for one return type and same goes with Array as well. So it could be either Promise<string[]> or Promise<Array<string[]>>.

packages/revocation/src/implementations/revocationOnChain.ts
```diff
@@ -0,0 +1,127 @@
+import {
+  Contract, ethers, Event, utils,
+} from 'ethers';
+import { EwSigner, addressOf } from '@ew-did-registry/did-ethr-resolver';
+import { abi as RevocationOnChainAbi } from '../constants/RevocationRegistryOnChain.json';
```
jrhender, line 5: I think it would be to include iam-contracts as a dependency and then could import the contract json from iam-contracts directly.

packages/revocation/src/implementations/revocationOnChain.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+  * @param owner - Entity which controls revocation
+  * @param addressOnchain - Address of the on chain claim's RevocationRegistry
+  */
+  constructor(
+    owner: EwSigner,
+    addressOnChain: string,
+  ) {
+    this._revocationRegistryOnChain = new ethers.Contract(
```
jrhender, line 18: Not absolutely necessary, but it could be nice to compile a Contract factory type using typechain to use here (see iam-client-lib package.json for example). What do you think of this?

packages/revocation/src/contracts/RevocationRegistry.sol
```diff
@@ -0,0 +1,31 @@
+pragma solidity >=0.7.0 <0.9.0;
+
+/**
+* A contract for tracking the revocations of off-chain verifiable credentials.
+* For on-chain role revocations, see contract in iam-contracts repository.
+*/  
+contract RevocationRegistry {
```
jrhender, line 7: Not exactly the right place for this comment, but I think that it is strange that `RevocationRegistryOffChain.json` is committed under `constants`. I think it would be better to compile the contract from the solidity as needed (see `proxyIdentity` package for example)

packages/revocation/src/implementations/revocationOnChain.ts
```diff
@@ -0,0 +1,127 @@
+import {
+  Contract, ethers, Event, utils,
+} from 'ethers';
+import { EwSigner, addressOf } from '@ew-did-registry/did-ethr-resolver';
+import { abi as RevocationOnChainAbi } from '../constants/RevocationRegistryOnChain.json';
+
+export class RevocationOnChain {
```
jrhender, line 7: Are there tests for this class? I don't see any at the moment

packages/revocation/src/implementations/revocationOffChain.ts
```diff
@@ -0,0 +1,62 @@
...
+      );
+      if (!event) return false;
+    } catch (error) {
+      throw new Error(error);
+    }
+    return true;
+  }
+
+  /**
+  * checks the revocation details for a credential
```
jrhender, line 48:
```suggestion
  * Checks the revocation details for a credential
```

jrhender (changes requested):
Thanks for the changes. Added a few more comments

packages/revocation/src/implementations/revocationOnChain.ts
nichonien: Those tests are part of the IAM contract library. 

https://github.com/energywebfoundation/iam-contracts/blob/feature/EDR-25-Revocation_registry/test/RoleDefinitionResolver.testSuite.ts

packages/revocation/src/implementations/revocationOnChain.ts
nichonien: Yeah, sounds good to me. Will do so.

nichonien: This change could be made at a later stage as the current version of `TS` in the `ew-did-registry` is `3.9.7` and it does create error when using `@typechain`. Once the `TS` version is updated, we would add the factory build for contracts.

packages/revocation/src/implementations/revocationOnChain.ts
```diff
@@ -0,0 +1,127 @@
...
+  * @param { string[] } subjects - DIDs for which the role is being revoked
+  * @param { string } revoker - DID of the revoker
+  * @returns Promise<boolean>
+  */
+  async revokeRoleforDIDs(role: string, subjects: string[], revoker: string):
+  Promise<boolean> {
+    const roleHash = utils.namehash(role);
+    const revokerAddress = addressOf(revoker);
+    const revocationSubjects = [];
+    for (let i = 0; i < subjects.length; i++) {
```
JGiter, line 70:
```suggestion
    for await (let i = 0; i < subjects.length; i++) {
```

packages/revocation/contracts/RevocationRegistry.sol
```diff
@@ -0,0 +1,31 @@
...
+*/  
+contract RevocationRegistry {
+    
+    struct RevokedClaim {
+        address revoker;
+        uint revokedTimestamp;
+    }
+    mapping (bytes32 => RevokedClaim[]) revokedClaims;
+ 
+    function revokeClaim(bytes32 credential) public  {
```
JGiter, line 15: Could you please document what credentials are?

nichonien: Would add a comment.

JGiter:
> > Is verification will be implemented separately?
> > Is it possible to provide one interface for verification regardless of role registration type?
> 
> For the off-chain revocations, the verifier could check all the revocations and then ask the revoker for presentation of his claim which gave it the authority to do so. While for the on-chain claims revocation the verifier could simply call the `isRevoked` method and get the revocation status.
> 
> IMO it isn't feasible to provide a single interface for verification. Also, keeping the revocation approach & verification separate for both on-chain claims and off-chain claims makes it more convenient to be used as per the requirement by different orgs.
> 
> Let me know your thoughts.

I would like revocation to require as little actions as possible from participants. In particular is it possible to verify revoker without awaiting presentation from him and possible subsequent presentations from revokers in chain?

packages/revocation/src/implementations/revocationOnChain.ts
nichonien: As discussed, it's better to have this change once the Typescript version is updated and relevant changes are done.

packages/revocation/src/implementations/revocationOnChain.ts
nichonien: This throws error and I guess it's not needed as the contract call already has `await`.

packages/revocation/README.md
```diff
@@ -0,0 +1,114 @@
+## Revocation Package
+Verifiable credential that are issued to a DID should be revocable by the authorised DID. The `Revocation` implementation achieves this requirement for both on-chain roles and off-chain arbitrary verifiable credential.
```
knzeng-e, line 2:
```suggestion
Verifiable credentials that are issued to a DID should be revocable by the authorised DID. The `Revocation` implementation achieves this requirement for both on-chain roles and off-chain arbitrary verifiable credentials.
```

packages/revocation/README.md
```diff
@@ -0,0 +1,114 @@
...
+> The goal of the EW-DID is to support claim issuance and revocation for DIDs on Energy Web Chain. The library uses two separate contracts for revoking on-chain roles and arbitrary verifiable credential respectively.
```
knzeng-e, line 3:
```suggestion
> The goal of the EW-DID is to support claim issuance and revocation for DIDs on Energy Web Chain. The library uses two separate contracts for revoking on-chain roles and arbitrary verifiable credentials respectively. 
```

knzeng-e (approved):

PR #338, opened 2021-09-30
Set token to enable ci write access on dev branch

knzeng-e:
### Summary

|             |   |
|-------------|---|
| Description |  When deploying on develop branch from github action, the workflow breaks due since this branch is a protected one which only accept merges after at least one review on Pull Request. This PR gives the github action direct write acess using a github token credited with this right |

#### Type of request

### List of Features

- Setting NPM Token
- Setting Github Token
- using Semantic-release-action@v

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -70,6 +70,7 @@ jobs:
             NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         
       - name: Commit to develop branch and merge to master branch
+        if: github.ref == 'refs/heads/master'
         run: |
           git checkout development
           git merge origin/master
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -56,10 +56,11 @@ jobs:
 
       - name: Deploy on development branch
         if: github.ref == 'refs/heads/development'
+        uses: codfish/semantic-release-action@v1
         env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_CI }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -56,10 +56,11 @@ jobs:
 
       - name: Deploy on development branch
         if: github.ref == 'refs/heads/development'
+        uses: codfish/semantic-release-action@v1
         env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_CI }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           RUN_NUMBER: ${{ github.run_number }}
-        run: |
-          npm run publish:canary
```
jrhender, line 62: Why the removal of `run publish:canary`? I guess the idea is that we are using `codfish` instead? Does `codfish` work with `lerna`?

knzeng-e: yes codfish should take care of publishing. I'm not sure how/if it works with lerna, on Codefish's [repository](https://github.com/codfish/semantic-release-action) they say that it

> respects any semantic-release configuration file in your repo or the release prop in your package.json.

knzeng-e: hum.. What should fix the permission issue isn't codfish but the token usage which will allow th ci action to deploy on behalf of the token I created with write access, as I understand it.
But I'm also sceptic on the syncing process with develop branch at the end of the action (since I don't use the token anymore in this part - hence my modification [cc2f66f](https://github.com/energywebfoundation/ew-did-registry/pull/338/commits/cc2f66faa3c7c21c3ba8c947e218e49b5a9e1bc4) )

PR #347, opened 2021-10-22
EDR-92/upgrade typescript

nichonien:
### Summary

|             |   |
|-------------|---|
| Description |  The `Typescript` version has been upgraded to match the version in `IAM-contracts` , this enables use of implementations in IAM-contract in `ew-did-registry`. |
| Jira issue  |  https://energyweb.atlassian.net/browse/EDR-92 |

#### Type of request


knzeng-e (dismissed):
Sounds good Ashish, thank you !

jrhender:
@nichonien could you fix the codacy errors if possible please?

packages/did-ethr-resolver/src/implementations/ewSigner.ts
```diff
@@ -120,9 +120,11 @@ export class EwSigner extends Signer {
       // Signer could be returned by not actually have an address
       // Best to check now and "fail fast"
       await signer.getAddress();
-    } catch (err) {
-      err.message = `Error instantiating EwSigner from ethers Provider. Provider must have signer with address. ${err.message}`;
-      throw err;
+    } catch (error) {
+      if (error instanceof Error) {
+        error.message = `Error instantiating EwSigner from ethers Provider. Provider must have signer with address. ${error.message}`;
+        throw new Error(error.message);
```
JGiter, line 126: I think error should be thrown regardless of `error` type

packages/did-resolver-interface/src/models/resolver.ts
```diff
@@ -42,30 +43,12 @@ export interface IServiceEndpoint {
   type: string;
   serviceEndpoint: string;
   description?: string;
-  validity: BigNumber;
-  block: number;
+  validity?: BigNumber;
+  block?: number;
```
JGiter, line 47: Could you please elaborate why these properties has become optional?

package.json
```diff
@@ -4,10 +4,10 @@
   "description": "The javascript library client that can be used to manage decentralised identities on energy web chains",
   "main": "index.js",
   "scripts": {
-    "test": "lerna run test --stream",
+    "test": "lerna run test",
     "test-rpc": "run-with-testrpc -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" --port 8544 --accounts 20 --networkId=9 --gasLimit=10000000 \"lerna run test\" ",
     "setup": "lerna bootstrap --hoist --no-ci",
-    "build": "rm -rf build docs packages/*/dist && lerna run compile",
+    "build": "rm -rf build docs packages/*/dist && lerna run compile && lerna run build",
```
JGiter, line 10: I think details of the build process should be contained in packages. Besides it is not clear from the first look what is difference between `build` and `compile`

packages/did-resolver-interface/src/models/resolver.ts
nichonien: The reason these are made optional was because, there are `deleted` in https://github.com/energywebfoundation/ew-did-registry/blob/56744f7ec6376f43868774d28796c210b4bbaf62/packages/did-ethr-resolver/src/functions/functions.ts#L379

With the `typescript` version update, it has become more strict and the `build` throws error 
"src/functions/functions.ts(385,14): error TS2790: The operand of a 'delete' operator must be optional."

packages/did-ethr-resolver/src/implementations/ewSigner.ts
nichonien: This was done, because the `TS` version upgrade threw error while build "Object is of type unknown". So it has to be given a type before using. And also `any` cannot be used now with latest TS version.

packages/claims/src/claimsVerifier/claimsVerifier.ts
```diff
@@ -1,4 +1,4 @@
-// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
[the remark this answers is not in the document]
nichonien: Agreed!

package.json
nichonien: I guess learn works on the repository level and to enable `typechain` build for `RevocationRegistry` I added  it to the main `package.json` to invoke the build script call.

packages/did-ethr-resolver/src/implementations/ewSigner.ts
JGiter: But what if `error` is not of type Error? Should method return normally?

packages/did-resolver-interface/src/models/resolver.ts
JGiter: I think that the actual error is how public key is constructed in `wrapDidDocument`. `validity` and `block` are in fact required fields

package.json
nichonien: Moved the build inside package!

packages/did-ethr-resolver/src/implementations/ewSigner.ts
nichonien: IMO, a standard error should be instance of `Error` always and if someone wants to have custom error then it would be a custom message with a new instance of the `Error`. Also we are only making standard calls using libraries so it should be fine and would appreciate your suggestion to replace this implementation.

packages/did-resolver-interface/src/models/resolver.ts
nichonien: Could you please elaborate more on the implementation side, why exactly we are deleting the values before pushing to the `didDocument`?

nichonien: Removed the delete property call from `wrapDpcument` method.

JGiter: validity and block are removed because they is no point to keep them in the document. This confusion appeared because `IPublicKey` is used for implementation (blockchain) details and for representation, which was not the best decision

JGiter (approved):

PR #349, opened 2021-10-25
feature/ICL-134 add support for P256 key

JGiter:
### Summary

|             |   |
|-------------|---|
| Description | Add support for ES256 jwt algorithm |
| Jira issue  | https://energyweb.atlassian.net/jira/software/projects/ICL/boards/85?selectedIssue=ICL-134 |

### List of changes
- jwt created from `Keys` or `ethers.Signer` are identified by non-standard `ES256K` method
- Added verification of `ES256` token
- `Algorithms` type moved from `did-resolver-interface` to `keys`, as it is type of key. Otherwise keys-specific functions would be also resided in `did-resolver-interface` or `keys` would depend on `did-resolver-interface`
- `noTimestamp` option of jwt verification is removed as it is makes no sense
- allowed empty catch block. It is needed to simplify jwt verification against several methods


packages/jwt/test/jwt.test.ts
```diff
@@ -1,80 +1,81 @@
...
+
+  describe('[ES256]', () => {
+    before(() => {
+      algorithm = 'ES256';
+      verifier = new JWT(new Keys());
+      const privKey = ECKey.createECKey('prime256v1');
+      signerPubKey = privKey.publicCodePoint.toString('hex');
+      token = jsonwebtoken.sign(payload, privKey.toString('pem'), { algorithm: 'ES256', noTimestamp: true });
+    });
+    signerPubKey
```
knzeng-e, line 50: Did you forget something here ? I don't get the purpose of this line

packages/jwt/test/jwt.test.ts
```diff
@@ -1,80 +1,81 @@
...
+    it('jwt signed with private key should return a string', async () => {
+      expect(token).to.be.a('string');
+    });
+
+    it('signed token should be verified', async () => {
+      expect(verifier.verify(token, signerPubKey, { algorithms: [algorithm] })).to.eql(payload);
+    });
+
+    it('verification with wrong signature should throw error', async () => {
+      return expect(function () { verifier.verify(token, new Keys().publicKey, { algorithms: [algorithm] }); }).to.throw();
```
knzeng-e, line 34: Maybe we can shorten a little bit the size of line ?
```suggestion
      expect(function () { verifier.verify(token, new Keys().publicKey, { algorithms: [algorithm] }); }).to.throw();
```

knzeng-e (dismissed):
That sounds good to me! Just added a few comments and suggestions.
Thanks @JGiter

packages/jwt/src/index.ts
```diff
@@ -131,11 +156,92 @@ class JWT implements IJWT {
...
+    throw new Error(INVALID_SIGNATURE);
+  };
+
+  private verifyJwt(token: string, publicKey: string, algorithm: Algorithms) {
+    const algorithms = [this.keyAlgToJwtAlg(algorithm)];
+    const pemPublicKey = pubToPem(publicKey, algorithm);
+    try {
+      return jsonwebtoken.verify(token, pemPublicKey, { algorithms });
+    } catch (_) {
+      throw new Error(INVALID_SIGNATURE);
```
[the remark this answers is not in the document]
JGiter: I've replaced error message with class, because it allows extend it with subclasses representing specific kind of such error

packages/revocation/src/constants/constants.ts
```diff
@@ -3,4 +3,4 @@ export const RevocationRegOffChainAddr = '0x3A859706c2491993Cf07760bf68F40B10946
 export const RevocationRegOnChainAddr = '0xb155d0ff939de83a5f26676fEcEd7f208C39055C';
 export const ENSRegistryAddr = '0xd7CeF70Ba7efc2035256d828d5287e2D285CD1ac';
 export const ClaimManagerAddr = '0xC3dD7ED75779b33F5Cfb709E0aB02b71fbFA3210';
-export const ERC1056Addr = '0x84d0c7284A869213CB047595d34d6044d9a7E14A';
\ No newline at end of file
+export const ERC1056Addr = '0x84d0c7284A869213CB047595d34d6044d9a7E14A';
```
jrhender, line 6: What changed on this line 🤔

packages/keys/src/models/index.ts
```diff
@@ -1,12 +1,14 @@
...
     PRIVATE_KEY: /^[a-f0-9]{64}$/,
     PUBLIC_KEY: /^[a-f0-9]{66}$/,
     SIGNATURE: /^[a-f0-9]{128}$/,
   },
 };
+
+/**
+ * Algorithms specifies, which algorithm has to be used with a particular public key
+ */
+export enum Algorithms {
```
jrhender, line 12: Can we change this to `KeyTypes`?
```suggestion
/**
 * KeyType specifies which curve (for elliptic curve keys) or algorithm (e.g. RSA) a key is associated with 
 */
export enum KeyType {
```

packages/keys/src/functions/index.ts
```diff
@@ -13,9 +18,39 @@ const hex = {
...
-  hex,
-  sha256,
+export const privToPem = (key: string, alg: Algorithms) => {
+  if (key.startsWith('0x')) {
+    key = key.slice(2);
+  }
+  return new ECKey({
+    crv: algToSecCrv(alg),
+    privateKey: Buffer.from(key, 'hex'),
+  }).toString('pem');
```
jrhender, line 55: I'm confused by these functions. Can't we combine them? It seems as though both are functionally the same?

packages/keys/src/functions/index.ts
```diff
@@ -13,9 +18,39 @@ const hex = {
...
+  } if (alg === Algorithms.Secp256r1) {
+    return 'P-256';
+  }
+  throw new Error(`Algorithm ${alg} is not supported`);
+};
+/**
+ *
+ * @param key compressed or uncompressed public key or private key in hexadecimal format
+ * @param alg
+ * @returns
```
jrhender, line 35: Missing documentation

packages/jwt/src/types.ts
```diff
@@ -0,0 +1,13 @@
+export type Algorithms = 'ES256' | 'ES256K';
```
jrhender, line 1: I still think we should change it from `ES256K` to something else. Maybe `EIP191` is the best as it describes what was performed: https://eips.ethereum.org/EIPS/eip-191
```suggestion
export type Algorithms = 'ES256' | 'EIP191';
```

jrhender (changes requested):

packages/revocation/src/constants/constants.ts
JGiter: It was changed by eslint, which requires empty line at the end of file

packages/keys/src/functions/index.ts
JGiter: It would complicate function logic, which depending on key type will strip or prepend `0x` and uncompress. Besides I am having problem with identifying var, which determines if key is private or public :smile: (alg is going to be renamed to keyType)

packages/jwt/src/JwtBase.ts
```diff
@@ -0,0 +1,100 @@
...
+  * @param {string} publicKey - public key in hexadecimal format in compressed or uncompressed form
+  */
+  verify(
+    token: string,
+    publicKey: string,
+    { algorithms = [Algorithms.EIP191] }: JwtVerifyOptions = {},
+  ): unknown {
+    if (algorithms.includes(Algorithms.EIP191)) {
+      try {
+        return this.verifyES256K(token, publicKey);
```
jrhender, line 40: I think it would be better if the `verify` function also reflected the Algo name
```suggestion
        return this.verifyEIP191(token, publicKey);
```

jrhender (changes requested):
One more change please 😄

jrhender (approved):
Looks good! Let's give it a try 👍

jrhender (approved):
Looks good!

PR #352, opened 2021-11-03
test(jwt): add test against bacem's token

jrhender:
### Summary
A test against the IoT safe JWT token


jrhender:
Closing as Bacem's token is working now

PR #355, opened 2021-11-17
Edr 88/support for ewc dids

nichonien:
### Summary

|             |   |
|-------------|---|
| Description | Add support for EWC & Volta DIDs, this would enable to resolve DID on respective chains  |
| Jira issue  |  https://energyweb.atlassian.net/browse/EDR-88 |

#### Type of request

- [x ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/packages/did-ethr-resolver/src/constants/constants.ts
+++ b/packages/did-ethr-resolver/src/constants/constants.ts
@@ -11,6 +11,6 @@ export const defaultProvider = {
 
 const pubKeyPattern = '0x[A-Fa-f0-9]{66}';
 export const attributeNamePattern = '^did/(pub|auth|svc)/(\\w+)(/(\\w+))?(/(\\w+))?$';
-export const delegatePubKeyIdPattern = `^did:[a-z0-9]+:${ethAddrPattern}#delegate-(sigAuth|veriKey)-(${pubKeyPattern}|${ethAddrPattern})$`;
-export const pubKeyIdPattern = `^did:[a-z0-9]+:${ethAddrPattern}#key-([A-Za-z0-9]+)`;
+export const delegatePubKeyIdPattern = `^did:[a-z0-9]+?(:)?[a-z0-9]+?:${ethAddrPattern}#delegate-(sigAuth|veriKey)-(${pubKeyPattern}|${ethAddrPattern})$`;
+export const pubKeyIdPattern = `^did:[a-z0-9]+?(:)?[a-z0-9]+?:${ethAddrPattern}#key-([A-Za-z0-9]+)`;
 export const serviceIdPattern = `^did:[a-z0-9]+:${ethAddrPattern}#service-([A-Za-z0-9]+)`;
--- a/packages/did-ethr-resolver/src/constants/constants.ts
+++ b/packages/did-ethr-resolver/src/constants/constants.ts
@@ -11,6 +11,7 @@ export const defaultProvider = {
 
 const pubKeyPattern = '0x[A-Fa-f0-9]{66}';
 export const attributeNamePattern = '^did/(pub|auth|svc)/(\\w+)(/(\\w+))?(/(\\w+))?$';
-export const delegatePubKeyIdPattern = `^did:[a-z0-9]+?(:)?[a-z0-9]+?:${ethAddrPattern}#delegate-(sigAuth|veriKey)-(${pubKeyPattern}|${ethAddrPattern})$`;
-export const pubKeyIdPattern = `^did:[a-z0-9]+?(:)?[a-z0-9]+?:${ethAddrPattern}#key-([A-Za-z0-9]+)`;
-export const serviceIdPattern = `^did:[a-z0-9]+:${ethAddrPattern}#service-([A-Za-z0-9]+)`;
+export const didPattern = `did:[a-z0-9]+?(:[a-z0-9]+)?:${ethAddrPattern}`;
+export const delegatePubKeyIdPattern = `^${didPattern}#delegate-(sigAuth|veriKey)-(${pubKeyPattern}|${ethAddrPattern})$`;
+export const pubKeyIdPattern = `^${didPattern}#key-([A-Za-z0-9]+)`;
+export const serviceIdPattern = `^${didPattern}#service-([A-Za-z0-9]+)`;
--- a/packages/did-ethr-resolver/src/functions/functions.ts
+++ b/packages/did-ethr-resolver/src/functions/functions.ts
@@ -17,6 +17,8 @@ import {
 } from '@ew-did-registry/did-resolver-interface';
 import { DIDPattern } from '@ew-did-registry/did';
 import { attributeNamePattern } from '../constants';
+// temporary import, to be imported from did package in future
+import { DIDPatternEWC, DIDPatternVOLTA } from '../utils';
 
 /**
  * This function updates the document if the event type is 'DelegateChange'
@@ -81,7 +83,18 @@ const handleAttributeChange = (
   validTo: BigNumber,
   block: number,
 ): IDIDLogData => {
-  let match = did.match(DIDPattern);
+  let match;
+  if (did.split(':').length > 3) {
+    if (did.includes('ewc')) {
+      match = did.match(DIDPatternEWC);
+    }
+    else if (did.includes('volta')) {
+      match = did.match(DIDPatternVOLTA);
+    }
+  }
+  else {
+    match = did.match(DIDPattern);
+  }
   if (!match) {
     throw new Error('Invalid DID');
   }
@@ -221,13 +234,13 @@ const getEventsFromBlock = (
   contractInterface: utils.Interface,
   address: string,
 ): Promise<unknown> => new Promise((resolve, reject) => {
-  const [, , identity] = did.split(':');
+  var identity = did.split(':').pop();
 
   provider.getLogs({
     address,
     fromBlock: block.toNumber(),
     toBlock: block.toNumber(),
-    topics: [null, `0x000000000000000000000000${identity.slice(2).toLowerCase()}`] as string[],
+    topics: [null, `0x000000000000000000000000${identity?.slice(2).toLowerCase()}`] as string[],
   }).then((log) => {
     const {
       name, args, signature, topic,
@@ -276,7 +289,7 @@ export const fetchDataFromEvents = async (
   provider: providers.Provider,
   selector?: DocumentSelector,
 ): Promise<void> => {
-  const [, , identity] = did.split(':');
+  var identity = did.split(':').pop();
   let nextBlock;
   let topBlock;
   try {
@@ -288,7 +301,7 @@ export const fetchDataFromEvents = async (
 
   if (nextBlock) {
     document.owner = await contract.owners(identity);
-  } else {
+  } else if (identity) {
     document.owner = identity;
   }
 
--- a/packages/did-ethr-resolver/src/functions/functions.ts
+++ b/packages/did-ethr-resolver/src/functions/functions.ts
@@ -15,10 +15,8 @@ import {
   DelegateChangedEvent,
   DidEventNames,
 } from '@ew-did-registry/did-resolver-interface';
-import { DIDPattern } from '@ew-did-registry/did';
 import { attributeNamePattern } from '../constants';
-// temporary import, to be imported from did package in future
-import { DIDPatternEWC, DIDPatternVOLTA } from '../utils';
+import { addressOf, matchDIDPattern } from '../utils';
 
 /**
  * This function updates the document if the event type is 'DelegateChange'
@@ -83,25 +81,11 @@ const handleAttributeChange = (
   validTo: BigNumber,
   block: number,
 ): IDIDLogData => {
-  let match;
-  if (did.split(':').length > 3) {
-    if (did.includes('ewc')) {
-      match = did.match(DIDPatternEWC);
-    }
-    else if (did.includes('volta')) {
-      match = did.match(DIDPatternVOLTA);
-    }
-  }
-  else {
-    match = did.match(DIDPattern);
-  }
-  if (!match) {
-    throw new Error('Invalid DID');
-  }
-  const identity = match[1];
+  const matchDID = matchDIDPattern(did);
+  const identity = matchDID[1];
   const attributeType = event.values.name;
   const stringAttributeType = ethers.utils.parseBytes32String(attributeType);
-  match = stringAttributeType.match(attributeNamePattern);
+  const match = stringAttributeType.match(attributeNamePattern);
   if (match) {
     const section = match[1];
     const algo = match[2];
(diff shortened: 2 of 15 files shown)


packages/did-ethr-resolver/src/implementations/resolver.ts
```diff
@@ -13,11 +13,18 @@ import {
...
+import {
+  Methods,
+  DIDPattern,
+  ethAddrPattern
+} from '@ew-did-registry/did';
 import { ethrReg } from '../constants';
 import { fetchDataFromEvents, wrapDidDocument, query } from '../functions';
 import { compressedSecp256k1KeyLength } from '..';
 
+// temporary declaration, will use as imported dependency from new release
```
JGiter, line 25: Is there is something prevents to move did-related code to did package now?

packages/did-ethr-resolver/src/implementations/resolver.ts
```diff
@@ -77,7 +84,18 @@ class Resolver implements IResolver {
     did: string,
     selector?: DocumentSelector,
   ): Promise<IDIDDocument> {
-    const match = did.match(DIDPattern);
+    let match;
```
JGiter, line 87: As extracting address from DID is common task I believe it make sense to create function and move it DID package

packages/did-ethr-resolver/src/functions/functions.ts
```diff
@@ -81,7 +83,18 @@ const handleAttributeChange = (
   validTo: BigNumber,
   block: number,
 ): IDIDLogData => {
-  let match = did.match(DIDPattern);
+  let match;
+  if (did.split(':').length > 3) {
```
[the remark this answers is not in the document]
nichonien: I'm not sure if I got your question, but this check is actually for validating both `ewc` and `Volta` did pattern based on the `split` length or else just the original `didPattern`.

packages/did-ethr-resolver/src/implementations/resolver.ts
nichonien: I was getting error after build if I tried importing it.

packages/did-ethr-resolver/src/implementations/resolver.ts
nichonien: You meant to say `match` `DIDPattern` is common task? If yes, that's a valid point and I will do the needful changes. I would use `AddressOf` method to extract address at places it is not being used. Thanks!

packages/did-ethr-resolver/src/functions/functions.ts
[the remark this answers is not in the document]
JGiter: I mean is it possible to validate ewc and volta DID's only by their patterns?

packages/did/src/index.ts
```diff
@@ -63,27 +63,37 @@ class DID implements IDID {
...
+      throw new Error('Network must not be empty and consist only of lowcase alphanumerical characters');
+    }
+    if (id !== undefined && !DID_SCHEME_PATTERNS.NETWORK.test(`${method}:${chain}`)) {
       throw new Error('Network must not be empty and consist only of lowcase alphanumerical characters');
     }
-    if (!DID_SCHEME_PATTERNS.ID.test(id)) {
+    if (id !== undefined && !DID_SCHEME_PATTERNS.ID.test(id)) {
+      throw new Error('Id must consist only of alphanumerical characters, dots, minuses and underscores');
+    }
+    if (id === undefined && !DID_SCHEME_PATTERNS.ID.test(chain)) {
```
jrhender, line 96: I'm finding this whole block of logic quite confusing. Could you explain it a bit more please?

Maybe to start, I think that `'DID must consist of three parts separated by a colon'` isn't true? We can have a did with 3 parts (`did:ethr:address`) or a DID with 4 parts (`did:ethr:ewc:address`), right?

Maybe this code would be clearer if first the length of the split DID was determined and then the variable names and error messages could be tailored to that?

jrhender (changes requested):
Some initial comments but it looks like a good start!

packages/did-ethr-resolver/src/utils/crypto.ts
```diff
@@ -1,12 +1,17 @@
...
+// importing Methods and ethAddrPattern temporarily to be used in did patterns
+import { DIDPattern, ethAddrPattern, Methods } from '@ew-did-registry/did';
 
 const {
   keccak256, hashMessage, arrayify, computePublicKey, recoverPublicKey, hexlify,
 } = utils;
 
+// temporary declaration, to be imported from @ew-did-registry/did after new publish
+export const DIDPatternEWC = `^did:${Methods.Erc1056}:ewc:(${ethAddrPattern})$`;
+export const DIDPatternVOLTA = `^did:${Methods.Erc1056}:volta:(${ethAddrPattern})$`;
```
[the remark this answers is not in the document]
nichonien: Yeah, It should be the case. But I was getting error post build, I might be missing something while importing or doing something wrong. A little more work might fix it.

packages/did-ethr-resolver/src/utils/crypto.ts
```diff
@@ -52,9 +57,45 @@ export function hexify(value: string | object): string {
...
+    if (did.includes('ewc')) {
+      match = did.match(DIDPatternEWC);
+    }
+    else if (did.includes('volta')) {
+      match = did.match(DIDPatternVOLTA);
+    }
+  }
+  else {
+    match = did.match(DIDPattern);
+  }
```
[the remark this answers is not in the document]
nichonien: Actually yes! There is no point in checking if its `ewc` or `volta` or anything. It should be a generalised pattern.

packages/did/src/index.ts
nichonien: Yeah, this looks confusing to me now when I randomly saw the logic. This definitely needs refactoring and right breakdown.

packages/did-ethr-resolver/src/constants/constants.ts
```diff
@@ -11,6 +11,6 @@ export const defaultProvider = {
 
 const pubKeyPattern = '0x[A-Fa-f0-9]{66}';
 export const attributeNamePattern = '^did/(pub|auth|svc)/(\\w+)(/(\\w+))?(/(\\w+))?$';
-export const delegatePubKeyIdPattern = `^did:[a-z0-9]+:${ethAddrPattern}#delegate-(sigAuth|veriKey)-(${pubKeyPattern}|${ethAddrPattern})$`;
-export const pubKeyIdPattern = `^did:[a-z0-9]+:${ethAddrPattern}#key-([A-Za-z0-9]+)`;
+export const delegatePubKeyIdPattern = `^did:[a-z0-9]+?(:)?[a-z0-9]+?:${ethAddrPattern}#delegate-(sigAuth|veriKey)-(${pubKeyPattern}|${ethAddrPattern})$`;
+export const pubKeyIdPattern = `^did:[a-z0-9]+?(:)?[a-z0-9]+?:${ethAddrPattern}#key-([A-Za-z0-9]+)`;
```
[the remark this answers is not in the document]
JGiter: May be it can be refactored even little more
```
const didPattern = `did:[a-z0-9]+?(:[a-z0-9]+)?:${ethAddrPattern}`;
```

JGiter (commented):
There is some DID handling code in resolver package, which from my point of view belongs to did package. I would create `parseDID`, which would return method and address when passed valid DID or null otherwise
Also if understand this task correctly resolver/operator should resolve document from chain depending on method

packages/did-ethr-resolver/src/functions/functions.ts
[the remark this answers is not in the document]
nichonien: We are just validating the did pattern only, nothing new has been added.

packages/did-ethr-resolver/src/functions/functions.ts
```diff
@@ -221,13 +218,13 @@ const getEventsFromBlock = (
...
 ): Promise<unknown> => new Promise((resolve, reject) => {
-  const [, , identity] = did.split(':');
+  var identity = addressOf(did);
 
   provider.getLogs({
     address,
     fromBlock: block.toNumber(),
     toBlock: block.toNumber(),
-    topics: [null, `0x000000000000000000000000${identity.slice(2).toLowerCase()}`] as string[],
+    topics: [null, `0x000000000000000000000000${identity?.slice(2).toLowerCase()}`] as string[],
```
jrhender, line 227: rather than `?` here, what do you think of throwing an error if `identity` is `undefined` as that should really never happen

packages/did-ethr-resolver/src/utils/crypto.ts
```diff
@@ -52,9 +52,23 @@ export function hexify(value: string | object): string {
...
     throw new Error('Invalid DID');
   }
   return match[1];
 }
+
+/**
+* Checks if did has a valid pattern, and returns the matched pattern array if it is
+*
+* @param did
+* @private
```
jrhender, line 66: Why this `@private` annotation?

packages/did/src/utils/parser.ts
```diff
@@ -0,0 +1,34 @@
+export const didPattern = `^did:[a-z0-9]+?:?[a-z0-9]+?:(0x[A-Fa-f0-9]{40})$`;
+
+/**
+* Checks if did is valid and returns the did method
+*
+* @param did
+* @private
```
jrhender, line 7: Why `@private` here? ts-doc says that it is discouraged as it may be removed: https://typedoc.org/guides/doccomments/#%40public%2C-%40protected%2C-and-%40private

packages/did/src/utils/parser.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+    const [, method, ...param] = did.split(':');
+    return method;
+}
+
+/**
+* Checks if did is valid and returns the chain did is associated with it
+*
+* @param did
+* @private
```
jrhender, line 22: Why `@private` here?

packages/did/src/utils/parser.ts
```diff
@@ -0,0 +1,34 @@
+export const didPattern = `^did:[a-z0-9]+?:?[a-z0-9]+?:(0x[A-Fa-f0-9]{40})$`;
+
+/**
+* Checks if did is valid and returns the did method
+*
+* @param did
+* @private
+*/
+export function getDIDMethod(did : string): string {
```
jrhender, line 9: Might as well return one of the `Methods` from `../models`, I think

packages/revocation/src/constants/constants.ts
```diff
@@ -3,4 +3,4 @@ export const RevocationRegOffChainAddr = '0x3A859706c2491993Cf07760bf68F40B10946
 export const RevocationRegOnChainAddr = '0xb155d0ff939de83a5f26676fEcEd7f208C39055C';
 export const ENSRegistryAddr = '0xd7CeF70Ba7efc2035256d828d5287e2D285CD1ac';
 export const ClaimManagerAddr = '0xC3dD7ED75779b33F5Cfb709E0aB02b71fbFA3210';
-export const ERC1056Addr = '0x84d0c7284A869213CB047595d34d6044d9a7E14A';
```
jrhender, line 6: @nichonien where is this address coming from?

packages/did-ethr-resolver/src/functions/functions.ts
```diff
@@ -276,7 +273,7 @@ export const fetchDataFromEvents = async (
   provider: providers.Provider,
   selector?: DocumentSelector,
 ): Promise<void> => {
-  const [, , identity] = did.split(':');
+  var identity = addressOf(did);
```
jrhender, line 276: why `var`?
```suggestion
  const identity = addressOf(did);
```

packages/did-ethr-resolver/src/functions/functions.ts
jrhender: Actually, it looks like `addressOf` is already throwing an error

jrhender (changes requested):
I think it's really coming along 🚀 ! Just a few more suggested changes

packages/revocation/src/constants/constants.ts
nichonien: In the initial days I got confused between `IdentityManager` and `DIDRegistry`, and so I had used IdentityManager's address in place of DIDRegistry's address.

packages/did-ethr-resolver/src/functions/functions.ts
```diff
@@ -221,13 +218,13 @@ const getEventsFromBlock = (
   contractInterface: utils.Interface,
   address: string,
 ): Promise<unknown> => new Promise((resolve, reject) => {
-  const [, , identity] = did.split(':');
+  var identity = addressOf(did);
```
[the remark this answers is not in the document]
nichonien: I don't really have anything to say, I might would have been day dreaming. 😂

packages/did-ethr-resolver/src/functions/functions.ts
nichonien: Actually `?` should be removed now, it was added when identity was obtained from the `split` and possibly could have been undefined.

packages/did-ethr-resolver/src/utils/crypto.ts
nichonien: I looked at [crypto.ts](https://github.com/energywebfoundation/ew-did-registry/blob/eb2ebbc922a6aad792018a1db14c8230de190748/packages/did-ethr-resolver/src/utils/crypto.ts#L52) and it was used, so I thought to follow the same pattern. I guess, we can remove all `private` annotation.

packages/did/src/utils/parser.ts
nichonien: Agreed, `private` annotations were used in [crypto.ts](https://github.com/energywebfoundation/ew-did-registry/blob/eb2ebbc922a6aad792018a1db14c8230de190748/packages/did-ethr-resolver/src/utils/crypto.ts#L52) so thought of following the same pattern, I will remove it from everywhere.

packages/did/src/utils/parser.ts
nichonien: Yeah, could be done. But then we will have to match all the methods available(we might have plenty of methods in future for our chains). I guess it's better to return just the `method` and the calling function could validate it at it's end.

packages/did/src/utils/parser.ts
nichonien: Would remove it!

packages/did/src/utils/parser.ts
jrhender: Ok, fair enough 👍

packages/did/src/utils/parser.ts
```diff
@@ -1,34 +1,32 @@
...
\ No newline at end of file
+  const match = did.match(didPattern);
+  if (!match) {
+    throw new Error('Invalid DID');
+  }
+  if (did.split(':').length > 3) {
+    const [, , chain, ...param] = did.split(':');
+    return chain;
+  }
+  return 'No chain info';
```
jrhender, line 31: I wonder if `undefined` or a result object like 
```
{
  foundChainInfo: boolean,
  chainInfo: string | undefined
}
```
Otherwise, clients will have to hardcode to look for `'No chain info'`, right?

jrhender (changes requested):
Sorry, I thought of one more suggestion 😅

jrhender (approved):
Looks good!

PR #356, opened 2021-11-19
feature/PDA-18 verification code consolidation

JGiter:
### Summary

|             |   |
|-------------|---|
| Description | Move authentication token verification from passport-did-auth to did-reg |
| Jira issue  | https://energyweb.atlassian.net/browse/PDA-18 |

### List of Features
- [x] `AuthTokenVerifier` moved to claims package to proof authentication and verification
- [x] authenticate issuers as part of claim verification

### List of bug-fixes
- Type of signer used to init JWT191 was determined by `instanceof` operator, which is not reliable, because prototype of the same type differs from version to version. Type checking will only include one property. For the `ethers.Signer` it will be `isSigner`, and for the `Keys` it is `privateKey` 
- replaced `no-shadow` rule of eslint by corresponding rule from @typescript-eslint, because of `error  'KeyType' is already declared in the upper scope` false error of eslint (https://github.com/typescript-eslint/tslint-to-eslint-config/issues/856) 
- stripping `validity` and `block` in `wrapDidDocument` is changed because of `unused variable` eslint error
- `Claims.verifySignature` is replaced by `ProofVerifier.verifyAssertionProof`

codacy checks failed, because some of its eslint rules differs from ours

### DIFF
--- a/packages/claims/src/claims/claims.ts
+++ b/packages/claims/src/claims/claims.ts
@@ -39,7 +39,7 @@ export class Claims implements IClaims {
   constructor(
     owner: EwSigner,
     protected document: IDIDDocumentFull,
-    protected store: IDidStore
+    protected store: IDidStore,
   ) {
     this.keys = { publicKey: owner.publicKey, privateKey: owner.privateKey };
     this.jwt = new JWT(owner);
@@ -76,7 +76,7 @@ export class Claims implements IClaims {
       issuerDoc?: IDIDDocument;
       holderDoc?: IDIDDocument;
       verificationPurpose?: VerificationPurpose;
-    } = {}
+    } = {},
   ): Promise<IPublicClaim | IPrivateClaim> {
     const token = await this.store.get(claimUrl);
     const claim = this.jwt.decode(token) as (IPublicClaim | IPrivateClaim) &
@@ -112,7 +112,7 @@ export class Claims implements IClaims {
 
   /**
    * @description Verifies that token stored at `claimUrl` is one of the services of `holderDoc`.
-   * 
+   *
    * @param claimUrl: url of the published claim
    * @param params.hashFns: The function used to determine the of hash of the claim
    * token used in the DID document. Used to verify that the DID document service endpoint
@@ -127,15 +127,15 @@ export class Claims implements IClaims {
     }: {
       hashFns?: { [alg: string]: (data: string) => string };
       holderDoc: IDIDDocument;
-    }
+    },
   ) {
     const token = await this.store.get(claimUrl);
     const service = holderDoc.service.find(
-      (s) => s.serviceEndpoint === claimUrl
+      (s) => s.serviceEndpoint === claimUrl,
     ) as IServiceEndpoint;
     if (!service) {
       throw new Error(
-        `No service endpoint found for ${claimUrl} in holder DID document`
+        `No service endpoint found for ${claimUrl} in holder DID document`,
       );
     }
     const { hash, hashAlg } = service;
--- a/packages/claims/src/claims/claims.ts
+++ b/packages/claims/src/claims/claims.ts
@@ -39,7 +39,7 @@ export class Claims implements IClaims {
   constructor(
     owner: EwSigner,
     protected document: IDIDDocumentFull,
-    protected store: IDidStore,
+    protected store: IDidStore
   ) {
     this.keys = { publicKey: owner.publicKey, privateKey: owner.privateKey };
     this.jwt = new JWT(owner);
@@ -48,10 +48,20 @@ export class Claims implements IClaims {
 
   /**
    * Verifies issuance and publishing of claim at `claimUrl`.
-   * On successful verification returns claim
+   * On success returns claim.
+   * Verification takes into account purpose on which claim was issued.
    *
-   * @param claimUrl {string}
-   * @param hashFns {{ [alg: string]: (data: string) => string }}
+   * @param claimUrl: Url of claim to be verified. This method will retrieve
+   * the claim using the didStore configured for the class
+   * @param params.hashFns: The function used to determine the of hash of the claim
+   * token used in the DID document. Used to verify that the DID document service endpoint
+   * matches the retrieved claim.
+   * @param params.issuerDoc: Document with verification methods to verify claim issuance with
+   * @param params.holderDoc: Document with service endpoints to verify claim publishing with
+   * @param params.verificationPurpose: Specifies which verification methods to use.
+   * `VerificationPurpose.Authenticate` is used to verify claim issued to authenticate identity
+   * `VerificationPurpose.Assertion` is used to assert issuer approval.
+   * By default verification asserts claim approval.
    *
    */
   async verify(
@@ -60,13 +70,13 @@ export class Claims implements IClaims {
       hashFns,
       issuerDoc,
       holderDoc,
-      verificationPurpose = VerificationPurpose.Authentication,
+      verificationPurpose = VerificationPurpose.Assertion,
     }: {
       hashFns?: { [alg: string]: (data: string) => string };
       issuerDoc?: IDIDDocument;
       holderDoc?: IDIDDocument;
       verificationPurpose?: VerificationPurpose;
-    } = {},
+    } = {}
   ): Promise<IPublicClaim | IPrivateClaim> {
     const token = await this.store.get(claimUrl);
     const claim = this.jwt.decode(token) as (IPublicClaim | IPrivateClaim) &
@@ -101,10 +111,13 @@ export class Claims implements IClaims {
   }
 
   /**
-   * @description Verifies that token stored at `claimUrl` represents service
-   * endpoint of `holderDoc`
-   * @param claimUrl
-   * @param param1
+   * @description Verifies that token stored at `claimUrl` is one of the services of `holderDoc`.
+   * 
+   * @param claimUrl: url of the published claim
+   * @param params.hashFns: The function used to determine the of hash of the claim
+   * token used in the DID document. Used to verify that the DID document service endpoint
+   * matches the retrieved claim.
+   * @param params.holderDod: Document in which to search for service endpoint.
    */
   async validateServiceEndpointToken(
     claimUrl: string,
@@ -114,14 +127,16 @@ export class Claims implements IClaims {
     }: {
       hashFns?: { [alg: string]: (data: string) => string };
       holderDoc: IDIDDocument;
-    },
+    }
   ) {
     const token = await this.store.get(claimUrl);
     const service = holderDoc.service.find(
-      (s) => s.serviceEndpoint === claimUrl,
+      (s) => s.serviceEndpoint === claimUrl
     ) as IServiceEndpoint;
     if (!service) {
-      throw new Error(`No service endpoint found for ${claimUrl} in holder DID document`);
+      throw new Error(
+        `No service endpoint found for ${claimUrl} in holder DID document`
+      );
     }
     const { hash, hashAlg } = service;
     const createHash = { ...hashes, ...hashFns }[hashAlg as string];
--- a/packages/claims/src/claims/claims.ts
+++ b/packages/claims/src/claims/claims.ts
@@ -146,7 +146,7 @@ export class Claims implements IClaims {
       (s) => s.serviceEndpoint === claimUrl,
     ) as IServiceEndpoint;
     if (!service) {
-      throw new Error(`No claim at ${claimUrl}`);
+      throw new Error(`No service endpoint found for ${claimUrl} in holder DID document`);
     }
     const { hash, hashAlg } = service;
     const createHash = { ...hashes, ...hashFns }[hashAlg as string];
--- a/packages/claims/src/claims/claims.ts
+++ b/packages/claims/src/claims/claims.ts
@@ -117,7 +117,7 @@ export class Claims implements IClaims {
    * @param params.hashFns: The function used to determine the of hash of the claim
    * token used in the DID document. Used to verify that the DID document service endpoint
    * matches the retrieved claim.
-   * @param params.holderDod: Document in which to search for service endpoint.
+   * @param params.holderDoc: Document in which to search for service endpoint.
    */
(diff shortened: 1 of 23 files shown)


packages/claims/src/claims/claims.ts
```diff
@@ -70,20 +73,30 @@ export class Claims implements IClaims {
...
-    if (!this.document.isValidDelegate(DelegateTypes.verification, claim.signer, claim.did)) {
-      throw new Error('Issuer isn\'t a use\'r delegate');
+    if (
+      !this.document.isValidDelegate(
+        DelegateTypes.verification,
+        claim.signer,
+        claim.did,
+      )
+    ) {
+      throw new Error("Issuer isn't a use'r delegate");
```
jrhender, line 94:
```suggestion
      throw new Error("Issuer isn't a valid user delegate");
```

packages/claims/src/claims/claims.ts
```diff
@@ -70,20 +73,30 @@ export class Claims implements IClaims {
...
+      iss: string;
+      sub: string;
+    };
     if (!(await this.verifySignature(token, claim.iss))) {
       throw new Error('Invalid signature');
     }
-    if (!this.document.isValidDelegate(DelegateTypes.verification, claim.signer, claim.did)) {
-      throw new Error('Issuer isn\'t a use\'r delegate');
+    if (
+      !this.document.isValidDelegate(
```
jrhender, line 88: In my mind, part of this story is to consolidate the verification code that is in this `verify` function with the `verify` function that is in `AuthTokenVerifier`. For instance, a crucial difference is that `AuthTokenVerifier.verify` allows for either `isIdentity` or `isDelegate` (i.e. `(await this.isIdentity(token)) || (await this.isDelegate(token))`), while this code is just `isDelegate` at the moment

packages/claims/src/claimsVerifier/claimsVerifier.ts
```diff
@@ -64,7 +64,7 @@ export class ClaimsVerifier extends Claims implements IClaimsVerifier {
           privateClaim.did,
         )
     ) {
-      throw new Error('Issuer isn\'t a use\'r delegate');
+      throw new Error("Issuer isn't a use'r delegate");
```
jrhender, line 67:
```suggestion
      throw new Error("Issuer isn't a user delegate");
```

jrhender (changes requested):
I think a bit more consolidation work is needed. But maybe we need to discuss a bit more what this entails 😄

packages/claims/src/claims/claims.ts
JGiter: @jrhender It makes sense to authenticate issuer in verification methods of `claims` package by `AuthTokenVerifier`. It will cause reading of document thought. Is this acceptable?

packages/claims/src/claims/claims.ts
```diff
@@ -63,27 +64,29 @@ export class Claims implements IClaims {
...
-   *  delegate and the authenticity of the issuer's signature
+   * Authenticates claim issuer and verifies that claim is present in subject document
    *
    * @param claimUrl {string}
    * @param hashFns {{ [alg: string]: (data: string) => string }}
    */
   async verify(
-    claimUrl: string, hashFns?: { [alg: string]: (data: string) => string },
+    claimUrl: string,
+    hashFns?: { [alg: string]: (data: string) => string },
```
jrhender, line 74:
```suggestion
    claimUrl: string,
    hashFns?: { [alg: string]: (data: string) => string },
    cachedIssuerDoc?: IDIDDocument
```

packages/claims/src/claims/claims.ts
```diff
@@ -63,27 +64,29 @@ export class Claims implements IClaims {
...
     const token = await this.store.get(claimUrl);
-    const claim = this.jwt.decode(token) as
-      (IPublicClaim | IPrivateClaim) & { iss: string; sub: string };
-    if (!(await this.verifySignature(token, claim.iss))) {
-      throw new Error('Invalid signature');
+    const claim = this.jwt.decode(token) as (IPublicClaim | IPrivateClaim) & {
+      iss: string;
+      sub: string;
+    };
+    const issuerDoc = await this.document.read(claim.iss);
```
jrhender, line 81:
```suggestion
    const issuerDoc = cachedIssuerDoc ?? await this.document.read(claim.iss);
```

packages/claims/src/claims/claims.ts
jrhender: This requires:
1. `validateCredentialServiceEndpoint`
```
function validateCredentialServiceEndpoint(token: string, hashFns, cachedHolderDIDDocument?) {
   const service = (await this.document.readAttribute(
      { service: { serviceEndpoint: claimUrl } },
      claim.sub,
    )) as IServiceEndpoint;
    const { hash, hashAlg } = service;
    const createHash = { ...hashes, ...hashFns }[hashAlg as string];
    if (hash !== createHash(token)) {
      throw new Error('Claim was changed');
    }
}
```
2. `decodeCredential(claim: Claim, cachedIssuerDoc?: IDIDDocument)`
```
/**
 * @param claimUrl can be an input that store API accepts
 */
function decodeCredential(claimUrl: string): Claim {
  const token = await this.store.get(claimUrl);
    const claim = this.jwt.decode(token) as (IPublicClaim | IPrivateClaim) & {
      iss: string;
      sub: string;
    };
}
```
2. `verifyClaim(claim: Claim, cachedIssuerDoc?: IDIDDocument)`

jrhender (commented):
Some further suggestions

packages/claims/src/claims/claims.ts
```diff
@@ -63,32 +67,66 @@ export class Claims implements IClaims {
...
+    const claim = this.jwt.decode(token) as (IPublicClaim | IPrivateClaim) &
+      JwtPayload;
+    if (!issuerDoc) {
+      issuerDoc = await this.document.read(claim.iss);
     }
-    if (!this.document.isValidDelegate(DelegateTypes.verification, claim.signer, claim.did)) {
-      throw new Error('Issuer isn\'t a use\'r delegate');
+    const authVerifier = new AuthTokenVerifier(issuerDoc);
+    if (await authVerifier.verify(token)) {
+      return claim;
```
jrhender, line 98: I'm not sure we can necessarily rely on the `AuthTokenVerifier` exclusively in this way as we can't be sure that the token is an "auth" token. Because you can have token signed for "authentication" or "verification" (at least in our current code)

Like we have this code in `operator.ts`
```
/**
 * The two main types of public keys are, in turn, verification key and signature authentication key
 */
export enum PubKeyType {
  SignatureAuthentication2018 = 'sigAuth', VerificationKey2018 = 'veriKey'
}
```

JGiter: Not sure how approach this. Claim issuer should be authenticated, but the claim is not always authentication claim. Should we change `AuthTokenVerifier` to authenticate claims signed by verification key?

packages/claims/src/claims/claims.ts
```diff
@@ -63,32 +67,71 @@ export class Claims implements IClaims {
...
-    claimUrl: string, hashFns?: { [alg: string]: (data: string) => string },
+    claimUrl: string,
+    {
+      hashFns,
+      issuerDoc,
+      holderDoc,
+    }: {
+      hashFns?: { [alg: string]: (data: string) => string };
+      issuerDoc?: IDIDDocument;
+      holderDoc?: IDIDDocument;
```
jrhender, line 84: This assumes a rename of `PubKeyType` to `PubKeyPurpose`
```suggestion
      hashFns,
      issuerDoc,
      holderDoc,
      verificationPurpose,
    }: {
      hashFns?: { [alg: string]: (data: string) => string };
      issuerDoc?: IDIDDocument;
      holderDoc?: IDIDDocument;
      verificationPurpose?: pubKeyRelationship
```

packages/claims/src/claims/claims.ts
```diff
@@ -63,32 +67,71 @@ export class Claims implements IClaims {
...
     }
-    const service = await this.document.readAttribute(
-      { service: { serviceEndpoint: claimUrl } }, (claim).sub,
+    await this.validateServiceEndpointToken(claimUrl, {
+      hashFns,
+      holderDoc,
+    });
+
+    const authVerifier = new AuthTokenVerifier(issuerDoc);
+    if (await authVerifier.verify(token)) {
```
jrhender, line 102:
```suggestion
    const tokenVerifier = new TokenVerifier(issuerDoc);
    if (await tokenVerifier.verify(token, verificationPurpose)) {
```

jrhender (commented):
A suggested way forward

jrhender (changes requested):
I think it looks good but I'm a bit concerned that there might be some duplication happening still. Maybe we don't need to fix right now and can make another story though. Let's discuss it.

packages/claims/src/claims/claims.ts
```diff
@@ -63,32 +72,86 @@ export class Claims implements IClaims {
...
+    switch (verificationPurpose) {
+      case VerificationPurpose.Authentication:
+        if (await tokenVerifier.authenticate(token)) {
+          return claim;
+        }
+        break;
+      case VerificationPurpose.Verification:
+        if (await tokenVerifier.verify(token)) {
+          return claim;
+        }
```
jrhender, line 120:
```suggestion
    const proofVerifier = new ProofVerifier(issuerDoc);
    switch (verificationPurpose) {
      case VerificationPurpose.Authentication:
        if (await proofVerifier.verifyAuthenticationProof(token)) {
          return claim;
        }
        break;
      case VerificationPurpose.Assertion:
        if (await tokenVerifier.verifyAssertionProof(token)) {
          return claim;
        }
```

jrhender (changes requested):
1. Remove `verifySignature` in `Claims` and `verifyPublicClaim` in `ClaimsUser`

packages/claims/src/claims/claims.ts
```diff
@@ -38,57 +47,89 @@ export class Claims implements IClaims {
...
-   * Verifies integrity of the claim, the claim is issued by the user
-   *  delegate and the authenticity of the issuer's signature
+   * Verifies issuance and publishing of claim at `claimUrl`.
+   * On success returns claim.
+   * Verification takes into account purpose on which claim was issued.
+   * Claim can be issued to authenticate identity or to assert issuer approval.
+   * By default asserts claim approval. 
    *
    * @param claimUrl {string}
    * @param hashFns {{ [alg: string]: (data: string) => string }}
```
jrhender, line 57: Could you update the documentation? I think we should add the new properties and I don't think we need the type in the docs 🤔  
```suggestion
   * @param claimUrl: Url of claim to be verified. This method will retrieve the claim using the didStore configured for the class
   * @param params.hashFns: The function used to determine the of hash of the claim token used in the DID document. Used to verify that the DID document service endpoint matches the retrieved claim.
   .... other params
```

jrhender (changes requested):
Need to add `isIdentity` to assertion verifier and a bit of documentation to improve

packages/claims/src/claimsVerifier/proofVerifier.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  /**
+   * @description checks that token is issued by identity verification delegate
+   *
+   * @param token
+   * @returns
+   */
+  public async verifyAssertionProof(token: string): Promise<string | null> {
+    if (await this.isVerificationDelegate(token)) {
```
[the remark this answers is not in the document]
JGiter: Issuing claim by key corresponding to address seems like authentication. May be instead LoginStrategy will verify with authentication purpose?

packages/claims/src/claims/claims.ts
```diff
@@ -30,65 +39,109 @@ export class Claims implements IClaims {
...
-   *
-   * @param claimUrl {string}
-   * @param hashFns {{ [alg: string]: (data: string) => string }}
+   * @description Verifies that token stored at `claimUrl` is one of the services of `holderDoc`.
+   * 
+   * @param claimUrl: url of the published claim
+   * @param params.hashFns: The function used to determine the of hash of the claim
+   * token used in the DID document. Used to verify that the DID document service endpoint
+   * matches the retrieved claim.
+   * @param params.holderDod: Document in which to search for service endpoint.
```
jrhender, line 120:
```suggestion
   * @param params.holderDoc: Document in which to search for service endpoint.
```

packages/claims/src/claimsIssuer/claimsIssuer.ts
```diff
@@ -63,7 +64,8 @@ export class ClaimsIssuer extends Claims implements IClaimsIssuer {
     const curve: sjcl.SjclEllipticalCurve = ecc.curves.k256;
     const g = curve.G;
     const claim: IPrivateClaim = this.jwt.decode(token) as IPrivateClaim;
-    if (!(await this.verifySignature(token, claim.iss as string))) {
+    const proofVerifier = new ProofVerifier(await this.document.read(claim.signer));
+    if (!(await proofVerifier.verifyAuthenticationProof(token))) {
```
jrhender, line 68: I think this should be `verifyAssertionProof` because I think a PrivateClaim is would usually be a credential of credential of some kind, not an authentication proof. If you look at the DIDDoc spec, issuing VCs is done by a key with an assertion relationship: https://www.w3.org/TR/did-core/#assertion
Did you have some other perspective here?
```suggestion
    if (!(await proofVerifier.verifyAssertionProof(token))) {
```

packages/claims/src/claimsVerifier/proofVerifier.ts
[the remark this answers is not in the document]
jrhender: No, I think that only should be able to sign both `authentication` and `verification` proofs using the `identity` key. We were already operating with the identity address's public key in Switchboard, so I think this isn't that controversial.

jrhender (changes requested):
Documentation looks great, thanks! I noticed one more issue in `issuerPrivateClaim`

jrhender (approved):
Looks good, thanks!

PR #364, opened 2021-12-08
Fix/ics 121/rewrap removes endpoints

jrhender:
### Summary

|             |   |
|-------------|---|
| Description | Making deep copy before deletion because otherwise logs in param were being affected, preventing repeat wrapping |
| Jira issue  |  https://energyweb.atlassian.net/browse/ICS-121  |

#### Type of request

### DIFF
--- a/packages/did-ethr-resolver/src/functions/functions.ts
+++ b/packages/did-ethr-resolver/src/functions/functions.ts
@@ -315,6 +315,34 @@ export const fetchDataFromEvents = async (
 };
 
 /**
+ * The logs from ERC1056 have a validity and a block number
+ */
+ interface ILogWithValidityAndBlock {
+  validity?: BigNumber
+  block?: number
+}
+
+/**
+ * Makes a copy of a log event and remove the validity and block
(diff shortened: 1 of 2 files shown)


nichonien (approved):
That's a nice approach @jrhender , great work!

JGiter (approved):
Thanks. Looks good. You might also want to correct `IPublicKey` type to not include `validity` and `block`

jrhender:
> Thanks. Looks good. You might also want to correct IPublicKey type to not include validity and block

@JGiter you mean `validity` and `block` should be removed from `IPublicKey`
```
export interface IPublicKey {
    id: string;
    type: string;
    controller: string;
    ethereumAddress?: string;
    publicKeyBase64?: string;
    publicKeyBase58?: string;
    publicKeyHex?: string;
    publicKeyPem?: string;
    publicKeyJwk?: string;
    publicKeyMultibase?: string;
    validity: BigNumber;
    block: number;
    [key: string]: string | number | BigNumber | undefined;
}
```

PR #403, opened 2022-02-03
task/IS-12 polyfill Node.js modules

JGiter:
### Summary

|             |   |
|-------------|---|
| Description | Build packages to use in Node.js and browser environment |
| Jira issue  | https://energyweb.atlassian.net/browse/IS-12 |

### changes
-  bundles will be consumed as `commonjs` libraries
-  npm dependencies are excluded from Node.js bundles. Browser bundles includes npm dependencies in order to polyfill their Node.js packages. This significantly increases bundle size. For example index.js of claims has 16 kb, but index.web.js has 1 Mb
- `revocation` package compiled by typescript as before, because bundling with webpack requires fixing of https://energyweb.atlassian.net/browse/ICL-240, which might not be needed since we are planning remove iam-contract dependency from ew-did-registry/revocation


jrhender:
So I'm understanding that with this current approach, there would be index.js and index.web.js for each package. Is this right?

JGiter:
> So I'm understanding that with this current approach, there would be `index.js` and `index.web.js` for each package. Is this right?

Correct. Though more commonly for browser bundle is `esm`. I'll rename it

JGiter:
> Gotcha, thanks. So does this mean that a browser would always use the `esm` file? So if iam-client-lib was used by a browser, it would use the `esm` version of `ew-did-registry` packages and if iam-client-lib was used in node, then the `js` version would be used?

Right. I tested importing did-reg packages in ionic. But we will also need to add polyfilles in `esm` bundle of iam-lib

jrhender (approved):
Looks good

PR #405, opened 2022-02-17
feat: update `isValidErc1056` to accept DID address with chain name

Harasz:
### Summary

|             |   |
|-------------|---|
| Description |  `isValidErc1056` assuming a valid DID address with and without method specific id |
| Jira issue  |   |

#### Type of request

### DIFF
--- a/packages/did/src/utils/validation.ts
+++ b/packages/did/src/utils/validation.ts
@@ -1,12 +1,17 @@
 import { Methods, Chain } from '../models';
 
 export const ethAddrPattern = '0x[A-Fa-f0-9]{40}';
-export const DIDPattern = `^did:${Methods.Erc1056}:(${ethAddrPattern})$`;
+export const DIDPattern = `^did:${Methods.Erc1056}:([A-Za-z0-9]+:)?(${ethAddrPattern})$`;
+export const DIDPatternMainnet = `^did:${Methods.Erc1056}:(${ethAddrPattern})$`;
 export const DIDPatternEWC = `^did:${Methods.Erc1056}:${Chain.EWC}:(${ethAddrPattern})$`;
 export const DIDPatternVOLTA = `^did:${Methods.Erc1056}:${Chain.VOLTA}:(${ethAddrPattern})$`;
 
 export function isValidErc1056(did: string): boolean {
   return new RegExp(DIDPattern).test(did);
+}
+
+export function isValidErc1056Mainnet(did: string): boolean {
+  return new RegExp(DIDPatternMainnet).test(did);
 }
 
 export function isValidErc1056EWC(did: string): boolean {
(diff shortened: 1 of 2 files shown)


packages/did/src/utils/validation.ts
```diff
@@ -1,7 +1,7 @@
 import { Methods, Chain } from '../models';
 
 export const ethAddrPattern = '0x[A-Fa-f0-9]{40}';
-export const DIDPattern = `^did:${Methods.Erc1056}:(${ethAddrPattern})$`;
+export const DIDPattern = `^did:${Methods.Erc1056}:([A-Za-z0-9]+:)?(${ethAddrPattern})$`;
```
JGiter, line 4:
```suggestion
export const DIDPattern = `^did:${Methods.Erc1056}:((${Chain.EWC|Chain.VOLTA}):)?(${ethAddrPattern})$`;
```

JGiter (changes requested):

packages/did/src/utils/validation.ts
nichonien: @jrhender we already have two separate methods [isValidErc1056EWC](https://github.com/energywebfoundation/ew-did-registry/blob/6799fe007c1dc9a1e0b43f973fb516938e545673/packages/did/src/utils/validation.ts#L12) and [isValidErc1056VOLTA](https://github.com/energywebfoundation/ew-did-registry/blob/6799fe007c1dc9a1e0b43f973fb516938e545673/packages/did/src/utils/validation.ts#L16) to validate `did` with specific `chain` name.

Harasz: We can also have two methods, one more generic to catch all Erc1056 methods and one just for EWF Erc1056 and it would be `isValidErc1056EWC || isValidErc1056VOLTA`. But since we need this for PDA, which should be network agnostic and not specific just for EWF project I believe, we shouldn't specify chain name in validator.

nichonien: IMO, we could have one general purpose `did` i.e. `did:ethr:address` pattern validator and two separate pattern validator for `ewc` and `volta`. This way we could limit the usability if `did`s to specific chain. So probably we do more strict validation for `did`s wherever needed.

Harasz: Check out my latest suggestion @jrhender @nichonien @JGiter

jrhender (changes requested):

packages/did/src/utils/validation.ts
JGiter: My opinion is that both `did:ethr:addr`, `did:ethr:chainId:addr` are valid DID according to DID spec. If some application doesn't support some chain, then it is responsibility of application to check chain-id-specifier with `parseDID`

jrhender: @JGiter  Is `parseDID` a function we have in `ew-did-registry`?

JGiter: @jrhender We have functions in did/utils/parser which return separately chain and method. We can replace them with `parseDID`

jrhender (approved):
Looks good to me!

JGiter (approved):

PR #409, opened 2022-03-22
task/ICL-261 update ethers

JGiter:
### Summary

|             |   |
|-------------|---|
| Description | iam-client-lib and its dependencies should use same `ethers` |
| Jira issue  | https://energyweb.atlassian.net/browse/ICL-261 |

### List of Changes

- `ethers` upgraded to the latest to ensure compatibility across iam stack
- due to issue with catching ganache events by `ethers` https://github.com/ethers-io/ethers.js/issues/615 tests, which relay on `provider.on` changed to fetch and parse transaction log


packages/revocation/package.json
```diff
@@ -35,7 +35,7 @@
     "@ew-did-registry/keys": "0.6.0",
     "@ew-did-registry/did-ethr-resolver": "0.6.0",
     "@energyweb/iam-contracts": "3.3.0",
-    "ethers": "^5.4.7"
+    "ethers": "5.4.7"
```
jrhender, line 38: Why not updating the version in the `revocation` package?

JGiter: There was some dependencies problem when I last time tried to update it. Ashish said we can leave it as it is because revocation anyways will be moved to `ew-credentials`. I'll try to update it though

jrhender (approved):

jrhender (dismissed):

jrhender (approved):
Looks good to me, thanks @JGiter !

PR #410, opened 2022-03-29
feat(credentials-interface): add `credentials-interface` package

Harasz:
### Summary

|             |   |
|-------------|---|
| Description | Add new package `@ew-did-registry/credentials-interface` that provides interfaces for credentials  |
| Jira issue  |  https://energyweb.atlassian.net/browse/EDR-109  |

#### Type of request

### List of Features

- add `credentials-interface` package


packages/credentials-interface/src/eip712proof.ts
```diff
@@ -0,0 +1,18 @@
...
+export interface EIP712Proof {
+  // ! This is a hack to get the type to work with PEX types. PEX is not supporting EIP712 signatures.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [x: string]: string | string[] | any;
+  '@context': string | string[];
+  type: string;
+  proofPurpose: string;
+  proofValue: string;
+  verificationMethod: string;
+  created: string;
```
jrhender, line 12: What about this?
```suggestion
export interface EIP712Proof extends IProof {
  // ! Need to add index signature to be compatible with PEX type
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  [x: string]: string | string[] | any;
```

jrhender: @Harasz Doesn't look like this is implemented yet. Did you try it and it doesn't work?

Harasz: Extending to IProof? I think it doesn’t make sense. This is interface just for eip712 proof, bc PEX doesn’t has one. Presentation and credential proof object has type Iproof | EIP712.

packages/credentials-interface/src/utils.ts
```diff
@@ -0,0 +1,100 @@
+import { Credential, VerifiableCredential } from './credentials.types';
+import { Presentation } from './presentation.types';
+
+function isObject(obj: unknown): obj is object {
+  return typeof obj === 'object' && obj !== null;
+}
+
+function hasOwnProperty<X extends object, Y extends PropertyKey>(
```
JGiter, line 8: Can we use https://lodash.com/docs/4.17.15#has?

packages/credentials-interface/src/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (!isObject(proof)) return false;
+  if (!hasRequiredProperties(proof, requiredProofProperties)) return false;
+  if (!hasOwnProperty(proof, 'jws') && !hasOwnProperty(proof, 'proofValue')) {
+    return false;
+  }
+
+  return true;
+}
+
+export function isCredential(
```
JGiter, line 41: This can be done more reliable with https://www.npmjs.com/package/ts-interface-checker

JGiter: Oh, sorry, it was asked already

packages/credentials-interface/src/utils.ts
Harasz: Probably yes, but why we need external library for such easy function?

packages/credentials-interface/src/utils.ts
Harasz: We have very simple type checking, do we really need library to do that?

packages/credentials-interface/src/utils.ts
JGiter: First it allows to reduce code. We wouldn't need to add `isObject`, `hasOwnedProperty`, `hasRequiredProperties` as they are all implemented in `lodash`. But most important is that it would allow to avoid some tricky errors. For example `isObject` returns true when it is passed function, which is probably not what we want

packages/credentials-interface/src/utils.ts
JGiter: Yes, it is completely optional and perhaps should be moved to separate task. But from my point of view it is worth efforts as it reduces amount of code, excludes errors and updates checker when type updates

packages/credentials-interface/src/utils.ts
Harasz: I agree that maybe it will prevent errors, but it won't reduce the code. I've created `hasOwnProperty` and `isObject` mainly for typescript reason (`obj is object`, `obj is X & Record<Y, unknown>`). These functions help me declare types and avoid `@ts-ignore`, bc input object is unknown.

packages/credentials-interface/src/utils.ts
Harasz: Okay, we can discuss this later on with other team

packages/credentials-interface/src/eip712proof.ts
jrhender: Hmm, well my rational for extending `IProof` is that I see EIP712Proof as being IProof but with additional properties so from a DRY standpoint, it doesn't make sense to duplicate the IProof props in EIP712Proof. Is it is IProof has props that aren't possible to have in EIP712Proof?

Even if extending, I would keep `credentialProof: IProof | EIP712` because to me that is saying that credentialProof can be either `IProof` or `EIP712Proof`, which is correct

Harasz: It will suggest that `EIP712Proof` can have a `jws` property because `IProof` has one. I think it can be misleading, don't you think?

packages/credentials-interface/src/utils.ts
JGiter: Do you mean that tsc doesn't compile? I thought that `@types/lodash` has definitions

packages/credentials-interface/src/eip712proof.ts
jrhender: Ok, good point. Let's keep as is then 👍

JGiter (approved):

jrhender (approved):

PR #412, opened 2022-04-12
task/279 credentials exchange types

JGiter:
### Summary

|             |   |
|-------------|---|
| Description |   |
| Jira issue  | https://energyweb.atlassian.net/browse/ICL-279 |

#### Type of request

### DIFF
--- a/packages/credentials-interface/src/exchange.types.ts
+++ b/packages/credentials-interface/src/exchange.types.ts
@@ -1,0 +1,79 @@
+export type ExchangeInvitation = {
+  type: string;
+  /**
+   * unique SSI session
+   */
+  url: string;
+};
+
+export const EWF_EXCHANGE_INVITATION =
+  'https://energyweb.org/out-of-band-invitation/vc-api-exchange';
+
+export type VpRequest = {
+  /**
+   * From https://w3c-ccg.github.io/vp-request-spec/#format :
+   * "Challenge that will be digitally signed in the authentication proof
+   *  that will be attached to the VerifiablePresentation response"
+   */
+  challenge: string;
+
+  query: VpRequestQuery[];
+
+  /**
+   * The schema for this property is taken from https://github.com/w3c-ccg/vc-api/issues/245
+   */
+  interact: { service: VpRequestInteractService[] };
+};
+
+/**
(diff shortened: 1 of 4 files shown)


nichonien (dismissed):

packages/credentials-interface/src/exchange.types.ts
```diff
@@ -0,0 +1,133 @@
+export type ExchangeInvitation = {
+  type: string;
+  body: {
+    /**
+     * exchange base url
+     */
+    url: string;
+  };
+};
```
jrhender, line 9: Sorry @JGiter I know this is the format that I put in the story. However, this was the "to be" format based on on https://energyweb.atlassian.net/browse/REBDEV-89. The actual current format is:
```
"presentationLink": {
    "type": "string",
    "url": "string"
  }
```
Maybe to keep ICL more flexible and avoid needing to change ew-did-registry if the format changes, we can keep `ExchangeInvitation` and `EWF_EXCHANGE_INVITATION` in ICL for now?

JGiter: no problem. from one side its make sense to put exchange types along with exchange logic (in iam-lib), but from the other these types are used in both iam-lib and ssi-hub. Besides it is potentially possible that there will be some relationships between exchange and credentials type

jrhender: Ah good point, maybe best to have the type in ew-did-registry then 👍

packages/credentials-interface/src/exchange.types.ts
```diff
@@ -0,0 +1,131 @@
...
+  limit_disclosure: 'required' | 'preferred';
+};
+
+export type Field = {
+  path: string[];
+  id?: string;
+  purpose?: string;
+  filter?: Filter;
+};
+
```
jrhender, line 108: I would consider putting these in a separate file (maybe called `presentation-exchange`) in order to group these types as they all come from the Presentation Exchange spec https://identity.foundation/presentation-exchange/

jrhender (approved):
Looks good, just an organizational comment

packages/credentials-interface/src/exchange.types.ts
JGiter: Ok. lets keep file size under control

PR #424, opened 2022-05-25
feat: add JWT expiration

Harasz:
### Summary

|             |   |
|-------------|---|
| Description | Add expiration to JWT |
| Jira issue  | https://energyweb.atlassian.net/browse/ICL-305  |

#### Type of request

### List of Features

- JWT expiration

### DIFF
--- a/packages/jwt/src/JWTEIP191.ts
+++ b/packages/jwt/src/JWTEIP191.ts
@@ -48,7 +48,7 @@ export class JWT extends JwtBase implements IJWT {
    */
   async sign(
     payload: string | JwtPayload,
-    { issuer, subject, noTimestamp }: JwtSignOptions = {}
+    { issuer, subject, noTimestamp, expirationTimestamp }: JwtSignOptions = {}
   ): Promise<string> {
     const header = {
       alg: 'ES256K',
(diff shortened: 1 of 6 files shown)


packages/jwt/src/JWTEIP191.ts
```diff
@@ -48,7 +48,7 @@ export class JWT extends JwtBase implements IJWT {
    */
   async sign(
     payload: string | JwtPayload,
-    { issuer, subject, noTimestamp }: JwtSignOptions = {}
+    { issuer, subject, noTimestamp, expirationTimestamp }: JwtSignOptions = {}
```
JGiter, line 51: There is also standard prop of JWT `expiresOn`. Could we use it?

jrhender (approved):
Looks good to me

packages/jwt/src/JWTEIP191.ts
Harasz: Do you mean `exp`? I think `expirationTimestamp` is more clear

JGiter: Yes, the standard name is `exp`. What I meant is property of `jsonwebtoken` https://www.npmjs.com/package/jsonwebtoken#jwtsignpayload-secretorprivatekey-options-callback. It should also validate expiration

Harasz: I see `expiresIn` but it expect different value. `jsonwbetoken` want a seconds after which token is expired, but I expect a timestamp after which token is expired. Can be misleading for some peoples if we agree on the same name convention

PR #425, opened 2022-05-27
task/ICL-309 status list types

JGiter:
### Summary

|             |   |
|-------------|---|
| Description | Add types to support status list in verifiable credential |
| Jira issue  |   |

#### Type of request

### DIFF
--- a/packages/credentials-interface/src/credentials.types.ts
+++ b/packages/credentials-interface/src/credentials.types.ts
@@ -1,4 +1,9 @@
-import { ICredentialSubject, IJsonLdCredential, IProof } from '@sphereon/pex';
+import {
+  ICredentialStatus,
+  ICredentialSubject,
+  IJsonLdCredential,
+  IProof,
+} from '@sphereon/pex';
 import { EIP712Proof } from './eip712proof.types';
 
 export type CredentialSubject = ICredentialSubject;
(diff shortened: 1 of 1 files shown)


nichonien (approved):

packages/credentials-interface/src/credentials.types.ts
```diff
@@ -1,11 +1,32 @@
...
+
+export enum CredentialStatusType {
+  StatusList2021Entry = 'StatusList2021Entry',
+}
+
+export enum CredentialStatusPurpose {
+  REVOCATION = 'revocation',
+}
+
+export interface StatusList2021Entry extends ICredentialStatus {
```
Harasz, line 25: `ICredentialStatus` interface require a `id`, in `StatusList2021Entry` we don't want that.

JGiter: I don't think this is proper use for never type. Besides since id is required `never` will not allow to ignore it

Harasz: This is a case that `never` was intruded. For example this is a Typescript implementation of exclude utility types `type Exclude<T, U> = T extends U ? never : T;`. We are extending object, so we can override every property that we want

Harasz: Sorry, we need `id` type, bc `DIDKit` is requiring it. So maybe id would be the as `statusListCredential`?

Harasz (approved):

packages/credentials-interface/src/credentials.types.ts
JGiter: specification doesn't allow `id` to be url of status list. perhaps it can be just uuid

Harasz: Can be `id` must be an URI - https://www.w3.org/TR/vc-data-model/#status. URL is a valid URI

PR #428, opened 2022-06-14
feat(statuslist2021): check status statuslist2021entry

nichonien:
### Summary

|             |   |
|-------------|---|
| Description |  Implement status check for StatusList2021 |
| Jira issue  | https://energyweb.atlassian.net/browse/EDR-116 |

- Since packages were setup in --no-ci mode often PR's include unrelated updates of package-lock.json, which was reason of 
https://github.com/energywebfoundation/ew-did-registry/runs/6929271416?check_suite_focus=true. To avoid this packages will be bootstraped in `ci` mode
- Supposedly sharing faucet among all tests leads to incorrect nonce. So far this is observed only locally, but we need to investigate this and serialize access to faucet or separate test setups


packages/revocation/src/implementations/statuslistentry-verification.ts
```diff
@@ -0,0 +1,100 @@
...
+import * as pako from 'pako';
+import axios from 'axios';
+
+/**
+ * A Class to verify the StatusList2021Entry
+ */
+export class StatusListEntryVerification {
+  private verify: (vc: string, proof_options: string) => Promise<any>;
+
+  constructor(
```
jrhender, line 15: IMO, some documentation on what `vc` and `proof_options` are would be very helpful.

packages/revocation/src/implementations/statuslistentry-verification.ts
```diff
@@ -0,0 +1,100 @@
...
+    callbackFunction: (vc: string, proof_options: string) => Promise<any>
```
jrhender, line 16: I think `callbackFunction` is too generic.
```suggestion
    vcProofVerificationCallback: (vc: string, proof_options: string) => Promise<any>
```

packages/revocation/src/implementations/statuslistentry-verification.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+    throw new Error('Not a https url');
+  }
+
+  /**
+   * Fetches the StatusListCredential
+   * @param statusListCredential URL to fetch from
+   * @returns
+   */
+  async fetchCredential(statusListCredentialUrl: string) {
```
jrhender, line 78: Can this be private?
```suggestion
  async private fetchCredential(statusListCredentialUrl: string) {
```

packages/revocation/src/implementations/statuslistentry-verification.ts
```diff
@@ -0,0 +1,100 @@
...
+    callbackFunction: (vc: string, proof_options: string) => Promise<any>
+  ) {
+    this.verify = callbackFunction;
+  }
+
+  /**
+   * Verifies the status of StatusListEntry
+   * @param statusListEntry which needs to be verified
+   */
+  async verifyStatusList(statusListEntry: StatusList2021Entry) {
```
jrhender, line 25: I think this documentation is a bit misleading. At this point, the function isn't verifying the status of a StatusListEntry. It is verifying whether or not the `StatusListCredential` is valid. One must also call `verifyEncodedListStatus` to check the entry status, right?
```suggestion
  /**
   * Verifies the status of StatusListEntry
   * @param statusListEntry which needs to be verified
   */
  async verifyStatusList(statusListEntry: StatusList2021Entry) {
```

packages/revocation/src/implementations/statuslistentry-verification.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Verifies status of the credential in the encodedList
+   * @param encodedList list with revoked status
+   * @returns true if the status is revoked
+   */
+  verifyEncodedListStatus(encodedList: string) {
+    const status = pako.ungzip(Buffer.from(encodedList, 'base64'));
+    return status[0] === 1;
```
jrhender, line 98: I don't think `0` should be hardcoded here. It is an unnecessary magic string.

packages/revocation/src/implementations/statuslistentry-verification.ts
jrhender: The name is also confusing. It is `verifyStatusList` but the param is `statusListEntry`. Should it be `verifyStatusListEntry`?

packages/revocation/src/implementations/statuslistentry-verification.ts
```diff
@@ -0,0 +1,100 @@
...
+  StatusList2021Entry,
+  VerifiableCredential,
+} from '@ew-did-registry/credentials-interface';
+import * as pako from 'pako';
+import axios from 'axios';
+
+/**
+ * A Class to verify the StatusList2021Entry
+ */
+export class StatusListEntryVerification {
```
jrhender, line 12: Could you document the expected API/usage of this class please? It seems to me like it would be as follows
```suggestion
/**
 * A Class to verify the StatusList2021Entry
 * Sample usage:
 *  const entryVerifier = new StatusListEntryVerification(vcProofVerificationCallback);
 *.  entryVerifier.verifyStatusList(statusListEntry);
 *.  const credential = entryVerifier.fetchCredential(statusListEntry.credentialStatus);
 *.  entryVerifier.verifyEncodedStatusList(credential);
 */
export class StatusListEntryVerification {
```
The above seems quite cumbersome, IMO. Can we simplify it to one API call? Just `entryVerifier.verifyStatusListEntry` that does everything?

jrhender (changes requested):
Good start but some further discussion/work needed IMO

packages/revocation/src/implementations/statuslistentry-verification.ts
jrhender: > So the idea is that one should only provide StatusListEntry from a VC and the API should verify everything (encodedList, StatuslListCredential, statusPurpose and so on). The verifyStatusList (should be verifyStatusListEntry) would be the main API which uses other sub methods within to completely verify or perform status check.

This makes sense to me, but let me know if you see it differently

package.json
```diff
@@ -5,8 +5,8 @@
   "main": "index.js",
   "scripts": {
     "test": "lerna run test",
-    "test-rpc": "run-with-testrpc -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" --port 8544 --accounts 20 --networkId=9 --gasLimit=10000000 \"lerna run test\" ",
-    "setup": "lerna bootstrap --hoist --no-ci",
+    "test-rpc": "run-with-testrpc -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" --port 8544 --accounts 20 --networkId=9 --gasLimit=10000000 -e 100000 \"lerna run test\" ",
+    "setup": "lerna bootstrap --hoist --ci",
```
Harasz, line 9: Whats the reason behind that change?

packages/credentials-interface/package.json
```diff
@@ -48,7 +48,7 @@
   },
   "dependencies": {
     "@ethersproject/abstract-signer": "5.6.0",
-    "@sphereon/pex": "1.1.0",
+    "@sphereon/pex": "^1.0.2",
```
Harasz, line 51: Downgrade? why?

packages/credentials-interface/src/credentials.types.ts
```diff
@@ -1,30 +1,31 @@
 import { ICredentialStatus, ICredentialSubject, IProof } from '@sphereon/pex';
-import { ICredential } from '@sphereon/pex/dist/main/lib/types/SSI.types';
+import { ICredential } from '@sphereon/pex/dist/main/lib/types';
+import { ICredentialContextType } from '@sphereon/pex/dist/main/lib/types/SSI.types';
 import { EIP712Proof } from './eip712proof.types';
 
-export type CredentialSubject = ICredentialSubject;
```
Harasz, line 5: We are using this type in ICL for example, let bring it back

packages/credentials-interface/src/status-list-credential.types.ts
```diff
@@ -0,0 +1,39 @@
...
+  id: string;
+  type: CredentialStatusType.StatusList2021;
+  statusPurpose: CredentialStatusPurpose;
+  encodedList: string;
+}
+
+export interface StatusList2021Credential
+  extends VerifiableCredential<StatusList2021> {}
+
+export const StatusList2021_Context = 'https://w3id.org/vc/status-list/2021/v1';
```
Harasz, line 24:
```suggestion
export const StatusList2021Context = 'https://w3id.org/vc/status-list/2021/v1';
```

packages/credentials-interface/src/status-list-credential.types.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const validateStatusListEntry = (
+  statusListEntry: StatusList2021Entry,
+  statusList: StatusList2021Credential
+) => {
+  if (statusListEntry.type !== CredentialStatusType.StatusList2021) {
+    throw new Error('Not a StatusList2021 type');
+  }
+  if (statusListEntry.statusPurpose !== CredentialStatusPurpose.REVOCATION) {
```
Harasz, line 33: I would add suspension check also

packages/did-ethr-resolver/test/did-operator.test.ts
```diff
@@ -483,6 +475,7 @@ describe('[RESOLVER PACKAGE]: DID-OPERATOR', function didOperatorTests() {
       address: registry,
     };
     operator = new Operator(owner, registrySettings);
+    console.log('> operator:', keys.getAddress());
```
Harasz, line 478:
```suggestion
```

packages/revocation/src/implementations/status-list-entry-verification.ts
```diff
@@ -0,0 +1,111 @@
...
+ * await statusListEntryVerifier.verifyStatusListEntry(statusListEntry);
+ * ```
+ */
+export class StatusListEntryVerification {
+  /**
+   *
+   * @param verifyProof function to verify proof of credential
+   */
+  constructor(
+    private verifyProof: (vc: string, proof_options: string) => Promise<any>
```
Harasz, line 33:
```suggestion
    private verifyProof: (vc: string, proofOptions: string) => Promise<unknown>
```

packages/revocation/src/implementations/status-list-entry-verification.ts
```diff
@@ -0,0 +1,111 @@
...
+  ) {}
+
+  /**
+   * Verifies that redential status exists, valid and has not been revoked
+   *
+   * @param credentialStatus link to credential status
+   */
+  async verifyCredentialStatus(
+    credentialStatus: StatusList2021Entry,
+    proof_options = JSON.stringify({})
```
Harasz, line 43:
```suggestion
    proofOptions = JSON.stringify({})
```

Harasz (changes requested):

packages/credentials-interface/src/credentials.types.ts
JGiter: But it has no value. Why not fix ICL?

packages/credentials-interface/src/status-list-credential.types.ts
JGiter: Hmm, may be you right. @jrhender Should this verification fail when credential status is suspended?

packages/credentials-interface/src/credentials.types.ts
Harasz: The main reason is to keep the same name convention. Pex is adding `I` to the beginning of type, we don't

JGiter: Okay

package.json
JGiter: In --no-ci mode  `npm run setup` updates packages dependencies. In result PR includes a lot of unrelated package-lock.json. In addition some of this updates breaks build as we has with ethers and recently with ipfs packages

Harasz: Okay, thanks

packages/credentials-interface/src/status-list-credential.types.ts
jrhender: Yes, I think it should also fail if suspended

JGiter: Ok, thanks

packages/credentials-interface/src/credentials.types.ts
```diff
@@ -1,30 +1,33 @@
 import { ICredentialStatus, ICredentialSubject, IProof } from '@sphereon/pex';
-import { ICredential } from '@sphereon/pex/dist/main/lib/types/SSI.types';
+import { ICredential } from '@sphereon/pex/dist/main/lib/types';
+import { ICredentialContextType } from '@sphereon/pex/dist/main/lib/types/SSI.types';
 import { EIP712Proof } from './eip712proof.types';
 
 export type CredentialSubject = ICredentialSubject;
 
-export interface Credential<T extends CredentialSubject> extends ICredential {
+export interface Credential<T extends ICredentialSubject> extends ICredential {
```
Harasz, line 8:
```suggestion
export interface Credential<T extends CredentialSubject> extends ICredential {
```

packages/credentials-interface/src/credentials.types.ts
```diff
@@ -1,30 +1,33 @@
...
+export enum CredentialStatusType {
+  StatusList2021 = 'StatusList2021',
+}
+
+export enum CredentialStatusPurpose {
+  REVOCATION = 'revocation',
+  SUSPENSION = 'suspension',
+}
+
+export interface VerifiableCredential<T extends ICredentialSubject>
```
Harasz, line 30:
```suggestion
export interface VerifiableCredential<T extends CredentialSubject>
```

packages/credentials-interface/src/status-list-credential.types.ts
```diff
@@ -0,0 +1,49 @@
...
+  encodedList: string;
+}
+
+// This interface is not empty because it is distinct from its generic parent
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface StatusList2021Credential
+  extends VerifiableCredential<StatusList2021> {}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Any = any;
```
Harasz, line 22: What is that and why?

Harasz (changes requested):
Couple of more changes and there are still old issues that weren't addressed

Harasz (dismissed):

Harasz (approved):

JGiter (approved):

PR #441, opened 2022-07-20
chore: update node version to 16.10

whitneypurdum:
### Summary

|             |   |
|-------------|---|
| Description | update node version to 16.10  |
| Jira issue  |  #122 |

#### Type of request

### List of Features

- feature #1
- feature #2


whitneypurdum:
To update the lockfile version, do we need to update the npm version to > 7? 

[package-lock.json | npm Docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion) 

[ew-did-registry/package.json at e44aa1b7001816c96433cda89a9eb69a36e45f75 · energywebfoundation/ew-did-registry](https://github.com/energywebfoundation/ew-did-registry/blob/e44aa1b7001816c96433cda89a9eb69a36e45f75/package.json#L84) 

I think this is the case, but am not positive. If so, I can add to open PR.

nichonien:
> To update the lockfile version, do we need to update the npm version to > 7?

Yeah, you need to have rpm version > 7 on your machine, this would create `package-lock.json` with version -2. 
> [ew-did-registry/package.json at e44aa1b7001816c96433cda89a9eb69a36e45f75 · energywebfoundation/ew-did-registry](https://github.com/energywebfoundation/ew-did-registry/blob/e44aa1b7001816c96433cda89a9eb69a36e45f75/package.json#L84)

Not sure why we have `npm` as a dependency in the `package.json`. @JGiter @jrhender any guesses why this is the case or if it's a mistake?

whitneypurdum:
> > To update the lockfile version, do we need to update the npm version to > 7?
> 
> Yeah, you need to have rpm version > 7 on your machine, this would create `package-lock.json` with version -2.
> 
> > [ew-did-registry/package.json at e44aa1b7001816c96433cda89a9eb69a36e45f75 · energywebfoundation/ew-did-registry](https://github.com/energywebfoundation/ew-did-registry/blob/e44aa1b7001816c96433cda89a9eb69a36e45f75/package.json#L84)
> 
> Not sure why we have `npm` as a dependency in the `package.json`. @JGiter @jrhender any guesses why this is the case or if it's a mistake?

I do have > 7 on my machine so I don't think that's the lockfileversion2 issue?
whitneypurdum@Whitneys-MacBook-Pro ew-did-registry % npm -v
8.5.5

nichonien:
> I do have > 7 on my machine so I don't think that's the lockfileversion2 issue? whitneypurdum@Whitneys-MacBook-Pro ew-did-registry % npm -v 8.5.5

What is the node version?

whitneypurdum:
> > I do have > 7 on my machine so I don't think that's the lockfileversion2 issue? whitneypurdum@Whitneys-MacBook-Pro ew-did-registry % npm -v 8.5.5
> 
> What is the node version?

whitneypurdum@Whitneys-MacBook-Pro ew-did-registry % node -v
v16.15.0

whitneypurdum:
Cancelling this workflow because the test commands runs without end after updating the version. Have retried the workflow twice and get same result. Perhaps a reason we did not update the node version before?

jrhender:
@whitneypurdum Maybe I'm missing something, but it looks like the test failed and wasn't necessarily hung. Is this new behaviour maybe?

https://github.com/energywebfoundation/ew-did-registry/runs/7452884309?check_suite_focus=true

>   1) [CREDENTIAL REVOCATION]
       "before each" hook for "A credential can be revoked multiple times":
     Error: processing response error (body="{\"id\":77,\"jsonrpc\":\"2.0\",\"error\":{\"message\":\"VM Exception while processing transaction: the tx doesn't have the correct nonce. account has nonce of: 37 tx has nonce of: 36 (vm hf=london -> block -> tx)\",\"stack\":\"

whitneypurdum:
> @whitneypurdum Maybe I'm missing something, but it looks like the test failed and wasn't necessarily hung. Is this new behaviour maybe?
> 
> https://github.com/energywebfoundation/ew-did-registry/runs/7452884309?check_suite_focus=true
> 
> > 1. [CREDENTIAL REVOCATION]
> >    "before each" hook for "A credential can be revoked multiple times":
> >    Error: processing response error (body="{"id":77,"jsonrpc":"2.0","error":{"message":"VM Exception while processing transaction: the tx doesn't have the correct nonce. account has nonce of: 37 tx has nonce of: 36 (vm hf=london -> block -> tx)","stack":"

Yes, new behavior. I'm not sure how to reset the nonce without a ui like metamask though.

nichonien:
@whitneypurdum I think we can close this.

PR #445, opened 2022-08-10
bug/SWTCH-1401 incorrect expiry

JGiter:
### Summary

|             |   |
|-------------|---|
| Description |   |
| Jira issue  |   |

#### Type of request

### DIFF
--- a/packages/jwt/src/JWTEIP191.ts
+++ b/packages/jwt/src/JWTEIP191.ts
@@ -72,7 +72,7 @@ export class JWT extends JwtBase implements IJWT {
       if (expirationTimestamp < Date.now()) {
         throw new Error('Expiration timestamp is in the past');
       }
-      payload.exp = Math.trunc(expirationTimestamp / 1000);
+      payload.exp = expirationTimestamp;
     }
     const encodedPayload = base64url(JSON.stringify(payload));
     const msg = `0x${Buffer.from(`${encodedHeader}.${encodedPayload}`).toString(
--- a/packages/claims/test/claimsIssuer.test.ts
+++ b/packages/claims/test/claimsIssuer.test.ts
@@ -116,7 +116,7 @@ describe('[CLAIMS PACKAGE/ISSUER CLAIMS]', function () {
       Buffer.from(payload, 'base64').toString('utf8')
     );
 
-    expect(decodedPayload.exp).to.eq(Math.trunc(expirationTimestamp / 1000));
+    expect(decodedPayload.exp).to.eq(expirationTimestamp);
   });
 
   it('claim issued by delegate should be verified', async () => {
--- a/packages/jwt/src/types.ts
+++ b/packages/jwt/src/types.ts
@@ -8,7 +8,10 @@ export type JwtSignOptions = {
   subject?: string;
   noTimestamp?: boolean;
   algorithm?: Algorithms;
-  expirationTimestamp?: number; // Number of milliseconds elapsed since January 1, 1970 00:00:00 UTC.
+  /**
+   * In milliseconds
+   */
+  expirationTimestamp?: number;
 };
 
 export type JwtVerifyOptions = {


packages/jwt/src/JWTEIP191.ts
```diff
@@ -72,7 +72,7 @@ export class JWT extends JwtBase implements IJWT {
       if (expirationTimestamp < Date.now()) {
         throw new Error('Expiration timestamp is in the past');
       }
-      payload.exp = Math.trunc(expirationTimestamp / 1000);
+      payload.exp = expirationTimestamp;
```
jrhender, line 75: @JGiter This change seems correct but I'm confused how credentials we were issuing previously had a correct expiry timestamp. Does this change require a corresponding change in callers of this code? Were callers compensating by multiplying the timestamp by 1000? Could you add some description to the PR about your understanding of this change please perhaps?

JGiter: @jrhender It is compensated in iam-lib here https://github.com/energywebfoundation/iam-client-lib/blob/6c74fb5c0d4185357b7678d47a5832126a9274a0/src/modules/claims/claims.service.ts#L1501. We will fix this in https://github.com/energywebfoundation/iam-client-lib/pull/633

jrhender: Hmm, interesting. That still doesn't explain how the timestamp is correct in previously issued JWTs as that line you linked in ICL to seems to be on the verification side. Does it make sense to you?

For example, @mhrsntrk issued this JWT yesterday using ICL and it seems to have the correct timestamp of 08/09/2023 11:22:28 (`1691569348010`)
```
eyJhbGciOiJFUzI1NksiLCJ0eXAiOiJKV1QifQ.eyJjbGFpbURhdGEiOnsicmVxdWVzdG9yRmllbGRzIjpbeyJrZXkiOiIxMjMiLCJ2YWx1ZSI6InRlc3QifV0sImNsYWltVHlwZSI6InJvbGUwMDMucm9sZXMuYXBwMDAxLmFwcHMubWhybmV4dGdlbi5pYW0uZXdjIiwiY2xhaW1UeXBlVmVyc2lvbiI6MSwiaXNzdWVyRmllbGRzIjpbXX0sImRpZCI6ImRpZDpldGhyOnZvbHRhOjB4ODU1NmM3MTRCQUI4MDFGZkI3MjkyQmQ3Mzk0MjQ2MGU3OWJlODNkRSIsInNpZ25lciI6ImRpZDpldGhyOnZvbHRhOjB4ODU1NmM3MTRCQUI4MDFGZkI3MjkyQmQ3Mzk0MjQ2MGU3OWJlODNkRSIsImNyZWRlbnRpYWxTdGF0dXMiOnsiaWQiOiJodHRwczovL2lkZW50aXR5Y2FjaGUtZGV2LmVuZXJneXdlYi5vcmcvdjEvc3RhdHVzLWxpc3QvdXJuOnV1aWQ6NmM2YmE5NWEtNDA1Mi00OGIzLTgwMTctMmNjZjMwM2Q1ODBmIiwidHlwZSI6IlN0YXR1c0xpc3QyMDIxRW50cnkiLCJzdGF0dXNMaXN0SW5kZXgiOiIwIiwic3RhdHVzTGlzdENyZWRlbnRpYWwiOiJodHRwczovL2lkZW50aXR5Y2FjaGUtZGV2LmVuZXJneXdlYi5vcmcvdjEvc3RhdHVzLWxpc3QvdXJuOnV1aWQ6NmM2YmE5NWEtNDA1Mi00OGIzLTgwMTctMmNjZjMwM2Q1ODBmIiwic3RhdHVzUHVycG9zZSI6InJldm9jYXRpb24ifSwiaXNzIjoiZGlkOmV0aHI6dm9sdGE6MHg4NTU2YzcxNEJBQjgwMUZmQjcyOTJCZDczOTQyNDYwZTc5YmU4M2RFIiwic3ViIjoiZGlkOmV0aHI6dm9sdGE6MHg4NTU2YzcxNEJBQjgwMUZmQjcyOTJCZDczOTQyNDYwZTc5YmU4M2RFIiwiZXhwIjoxNjkxNTY5MzQ4fQ.MHg4MTRiNTQzOTAwMTBmMTM4YjJjMzMwYzNkODA0NThhYmM0MTdmZDQ3ZmYzNmU3YjQzMmVkM2JkNzdhNjk2NjU2MjJhMjU0OGNkMzc4ZmI4NGUyYjA2ZmI5Nzk2YjNjMWU3MzNiY2Q5MDhmNjAwNjU2N2U3ZjI3MTI1MjcwNmZlNDFj
```

JGiter: It doesn't seems correct to me. The `1691569348010` is 1000 times greater then timestamp in jwt which is `1691569348`

jrhender: @JGiter Ah, it seems like `exp` should be in seconds, not milliseconds. So I think the current behaviour is correct and we should NOT merge this change https://stackoverflow.com/a/39926886/

jrhender (changes requested):
I believe this PR should not be merged as is and the current behaviour is correct.
We should perhaps add a link to the JWT RFC documentation to avoid confusion in the future.

packages/jwt/src/JWTEIP191.ts
JGiter: Agree. I'll revert it. Should we instead set expirationTimestamp in seconds?

jrhender: I don't think so, as this would be a breaking change to the library

PR #447, opened 2022-08-29
task/EDR-128 improve tests reliability

JGiter:
#### Type of request

### List of Features

- random signer in tests
- updated Node.js to 16

Sometimes `incorrect nonce` error still happens, but now it is more rarely, so this PR in any case improves situation. Besides test duration decresed to 2 minutes


jrhender:
I'm okay with this approach, but I think @nichonien had recently aligned the signer in tests (rather than a random signer) so I'm curious what his thoughts on this PR are

JGiter:
> I'm okay with this approach, but I think @nichonien had recently aligned the signer in tests (rather than a random signer) so I'm curious what his thoughts on this PR are

I think he mentioned signer in revocation package. Changing it more difficult because along such change requires also changing of issuance date, verification method and issuer in the credential should be changed

JGiter:
> Thanks @JGiter ! Although I'm curious to know how the node upgrade worked for this PR?

It is not really relates to this issue, but it won't hurt to align with other packages

PR #448, opened 2022-08-30
Update ethers 5.7

JGiter:
### Summary

|             |   |
|-------------|---|
| Description |   |
| Jira issue  |   |

### List of changes

- `incorrect account` error fixed by running ganache synchronously
- `ethers` updated to latest
- Node.js in `build` action wan't updated, but locally tests are passing on Node 18
- test accounts will be replenished when created, because these steps are always follows each other


.github/workflows/deploy.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           token: ${{ secrets.GH_TOKEN }}
       - uses: actions/setup-node@v2-beta
         with:
-          node-version: '16.10.0'
+          node-version: '18.6.0'
```
jrhender, line 34: Might as well use the latest I think
```suggestion
          node-version: '18.8.0'
```

jrhender (commented):
I think it would be better if this were split into two PRs: one that made the ethers changes and one that adjusted the tests. It seems to me these two changes aren't directly related and so it makes the PR harder to review and it would make it harder to revert just the ethers version update.

JGiter:
@jrhender `build` is failed on Node.js 18 although locally it is passing. May be github action support for non-LTS Node isn't complete. Another possible reason is abandonded `run-with-testprc` and deprecated `ganache-cli`. May be we should try `hardhat` with `ganache` https://hardhat.org/hardhat-runner/docs/other-guides/ganache-tests?

jrhender:
@JGiter I think we don't technically have to use node v18 for GHA. I think we can just update the ethers version and keep GHA at node v16. Would that work?

JGiter:
> @JGiter I think we don't technically have to use node v18 for GHA. I think we can just update the ethers version and keep GHA at node v16. Would that work?

Yes, it is works on 16. It just would be more reliable to test on actual environment

jrhender (approved):
Looks good to me, thanks @JGiter !

PR #451, opened 2022-09-20
task/SWTCH-1502 hosted ipfs cluster

JGiter:
### Summary

|             |   |
|-------------|---|
| Description | DidStore will persist claims in ipfs cluster https://ipfscluster.io/ |
| Jira issue  |  |


packages/did-ipfs-store/src/didStore.ts
```diff
@@ -1,46 +1,57 @@
...
+   * @param url Ipfs cluster root
+   * @param headers Http request headers
    */
-  constructor(uri: string | object) {
-    this.ipfs = ipfsClient(uri);
-  }
+  constructor(
+    private url: string,
+    private headers: Record<string, string> = {}
+  ) {}
```
nichonien, line 21: @JGiter Can you please add some method documentation here?

packages/did-ipfs-store/src/didStore.ts
```diff
@@ -1,46 +1,57 @@
...
-    return claim;
+  async get(cid: string): Promise<string> {
+    const { data: content } = await axios.get(`${this.url}/ipfs/${cid}`, {
+      headers: this.headers,
+    });
+    return Buffer.from(content).toString();
   }
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async delete(uri: string): Promise<boolean> {
```
nichonien, line 40: Some documentation here as well.

packages/did-ipfs-store/src/didStore.ts
JGiter: I can add more info about class itself. Public methods are documented in implemented interface

JGiter: Or do you think there something more can be said?

nichonien: Actually `get` method had some documentation while `save` and `delete` didn't have, so it could be consistent if we added some documentation for these two as well.

JGiter: Agree on consistency. Moved method documentation to method definitions

nichonien (dismissed):
Great work, thanks @JGiter.

jrhender (changes requested):
Thanks @JGiter , good work so far.

Some notes:
- Please update the "test" documentation to say that Docker now needs to be running: https://github.com/energywebfoundation/ew-did-registry#test
- As you noted, we should probably NOT merge this until we are ready from the infrastructure side
- I think this should be marked as a breaking change, as anyone who was using non-IPFS cluster as their IPFS infrastructure will no longer be able to use `did-ipfs-store` without migration, right? Or will this new `did-ipfs-store` be backwards compatible with non-cluster IPFS infrastructure?

README.md
```diff
@@ -42,6 +42,9 @@ npm run build
 ```
 
 ### Test
+
+Install docker-compose
+
```
jrhender, line 47: I think we need to tell people to actually run the docker-compose, right?
```suggestion
Install docker-compose

```bash
docker compose -f tests/setUpIpfs/docker-compose.yml
```

jrhender (commented):
Is this still missing the "BREAKING CHANGE" annotation in the commits? Or did I just miss it?

README.md
JGiter: I am not sure. docker-compose should be available right after installation. Or do you mean to start ipfs-cluster container? It is started in script https://github.com/energywebfoundation/ew-did-registry/blob/c81b0398da329cd463d007544c209d62321a2411/tests/setUpIpfs/index.ts#L15

JGiter:
> Is this still missing the "BREAKING CHANGE" annotation in the commits? Or did I just miss it?

No, sorry. I missed this comment

README.md
jrhender, line 47: Can we remove this text now that `setupIpfs` script is changed?

JGiter: May be to change it to `Install docker`?

jrhender (approved):
Looks good, thanks @JGiter

PR #456, opened 2022-10-11
check claim pin status

JGiter:
### Summary

|             |   |
|-------------|---|
| Description |  |
| Jira issue  |  |

### DIFF
--- a/packages/did-ipfs-store/src/didStore.ts
+++ b/packages/did-ipfs-store/src/didStore.ts
@@ -1,12 +1,18 @@
 import { IDidStore } from '@ew-did-registry/did-store-interface';
 import fetch from '@web-std/fetch';
 import { FormData } from '@web-std/form-data';
-import { File, Blob } from '@web-std/file';
+import { Blob } from '@web-std/file';
 import axios from 'axios';
 import { values } from 'lodash';
-import { StatusResponse, TrackerStatus } from './ipfs.types';
-
-Object.assign(global, { fetch, File, Blob, FormData });
+import {
+  AddResponse,
+  PinResponse,
+  StatusResponse,
+  TrackerStatus,
+  PIN_TIMEOUT,
+  REPLICATION,
+} from './ipfs.types';
+import { waitFor } from './utils';
 
 /**
  * Implements decentralized storage in IPFS cluster. Storing data in cluster allows to provide required degree of data availability
@@ -29,21 +35,22 @@ export class DidStore implements IDidStore {
 
   /**
    * @param claim stringified claim. Supported types of claim content are `string` and `object`
-   * @param pinTimeout Defines how to long to wait for pinning completion before throwing error, seconds
+   * @param waitForPinned checker of claim availability. Replication of claim requires time. This parameter allows to configure degree of availability after `save` returns. By default `save` waits until claim is pinned on `replication_factor_min` nodes
    */
-  async save(claim: string, pinTimeout = 5): Promise<string> {
+  async save(
+    claim: string,
+    waitForPinned?: (cid: string) => Promise<void>
+  ): Promise<string> {
     const blob = new Blob([claim]);
-    const { cid } = await this.add(blob);
+    let { cid } = await this.add(blob);
+    cid = cid.toString();
 
-    const { default: waitFor } = await (eval('import(`p-wait-for`)') as Promise<
-      typeof import('p-wait-for')
-    >);
+    if (!waitForPinned) {
+      waitForPinned = async (cid) => await this.waitForPinned(cid);
+    }
+    await waitForPinned(cid);
 
-    await waitFor(async () => await this.isPinned(cid), {
-      timeout: { milliseconds: pinTimeout * 1000 },
-    });
-
-    return cid.toString();
+    return cid;
   }
 
   async get(cid: string): Promise<string> {
@@ -64,10 +71,22 @@ export class DidStore implements IDidStore {
   /**
    * Checks if file is pinned on cluster
    * @param cid CID
+   * @param replicationFactor specifies on how many nodes data should be pinned https://ipfscluster.io/documentation/guides/pinning/#replication-factors
    */
-  async isPinned(cid: string): Promise<boolean> {
-    return values((await this.status(cid)).peer_map).some(
-      (pinInfo) => pinInfo.status === TrackerStatus.Pinned
+  async isPinned(cid: string, replicationFactor = REPLICATION.MIN): Promise<boolean> {
+    const { replication_factor_min, replication_factor_max } =
+      await this.allocations(cid);
+    const expectedReplicationFactor =
+      replicationFactor === REPLICATION.LOCAL
+        ? 1
+        : replicationFactor === REPLICATION.MIN
+        ? replication_factor_min
+        : replication_factor_max;
+
+    return (
+      values((await this.status(cid)).peer_map).filter(
+        (pinInfo) => pinInfo.status === TrackerStatus.Pinned
+      ).length >= expectedReplicationFactor
     );
   }
 
@@ -78,7 +97,7 @@ export class DidStore implements IDidStore {
   async status(cid: string): Promise<StatusResponse> {
     const path = `pins/${cid}`;
 
-    const data = await this.request(path, {
+    const data = await this.request<StatusResponse>(path, {
       method: 'GET',
       params: this.params,
     });
@@ -95,11 +114,12 @@ export class DidStore implements IDidStore {
     const body = new FormData();
     body.append('file', file);
     try {
-      const result = await this.request('add', {
+      const result = await this.request<AddResponse[]>('add', {
         method: 'POST',
         body,
         params: this.params,
       });
+
       const data = result[0];
       return { ...data, cid: data.cid };
     } catch (err) {
@@ -114,7 +134,13 @@ export class DidStore implements IDidStore {
     }
   }
 
-  private async request(
+  private async waitForPinned(cid: string): Promise<void> {
+    await waitFor(async () => await this.isPinned(cid), {
+      timeout: PIN_TIMEOUT * 1000,
+    });
+  }
+
+  private async request<T>(
     path: string,
     {
       method,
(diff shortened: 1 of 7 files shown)


nichonien (approved):

packages/did-ipfs-store/src/didStore.ts
```diff
@@ -27,9 +27,22 @@ export class DidStore implements IDidStore {
...
     const { cid } = await this.add(blob);
+
+    const { default: waitFor } = await (eval('import(`p-wait-for`)') as Promise<
+      typeof import('p-wait-for')
+    >);
+
+    await waitFor(async () => await this.isPinned(cid), {
+      timeout: { milliseconds: pinTimeout * 1000 },
+    });
+
```
jrhender, line 45: @JGiter could you explain a little more why this is necessary please? 

From the IPFS Cluster documentation, it is a bit unclear to me what the consistency semantics of the `add` operation are and how that aligns with the consistency semantics that this `did-store` `save` operation promises.
The [IPFS Cluster](https://ipfscluster.io/documentation/guides/pinning/#adding-files) documentation says:
> This means that when the add process is finished, your file will have been fully added to several IPFS daemons (and not necessarily the local one). 

So to me, this suggests that when `add` returns, one can be sure that IPFS Cluster has received the data (and is propagating it).

Maybe it is better to therefore have this `save` method return when the `add` method returns? In this way, the caller of `save` method could call it again before needing to wait for the full propagation to each node to be complete? 

Maybe as you have it is better, but just trying to understanding the rational a bit better 😃

packages/did-ipfs-store/src/didStore.ts
```diff
@@ -48,6 +61,30 @@ export class DidStore implements IDidStore {
...
   }
 
+  /**
+   * Checks if file is pinned on cluster
+   * @param cid CID
+   */
+  async isPinned(cid: string): Promise<boolean> {
+    return values((await this.status(cid)).peer_map).some(
+      (pinInfo) => pinInfo.status === TrackerStatus.Pinned
+    );
```
jrhender, line 71: How does this pin check work? If only checking `some`, is it possible that it would only be pinned on one node, but not the others? Is that the desired behaviour?

jrhender (commented):
Some initial comments

packages/did-ipfs-store/src/didStore.ts
JGiter: This delay was added in orded to ensure the behavior you quoted from IPFS Cluster documentation. When I tested DidStore locally data were available immediately after `add` returns, but during testing in Github Action pin status after `add` was `pinning`. May be this is because implementation of `nftstorage/ipfs-cluster` on which our implementation is based, is work in progress now

packages/did-ipfs-store/src/didStore.ts
JGiter: Yes, it is possible that data is pinned only on one node, because it is configured by replication factor, which is set in cluster configuration. Probably we can improve this part by waiting enough time to replicate on `replication_factor_min` nodes https://ipfscluster.io/documentation/guides/pinning/#replication-factors. I'll try this. Thanks for notice
It is also possible to set replication factor on adding, but I am not sure if we should expose this option as it requires some knowledge about this process

packages/did-ipfs-store/src/didStore.ts
```diff
@@ -27,9 +33,22 @@ export class DidStore implements IDidStore {
...
-  async save(claim: string): Promise<string> {
+  /**
+   * @param claim stringified claim. Supported types of claim content are `string` and `object`
+   * @param pinTimeout Defines how to long to wait for pinning completion before throwing error, seconds
+   */
+  async save(claim: string, pinTimeout = 5): Promise<string> {
     const blob = new Blob([claim]);
     const { cid } = await this.add(blob);
+
+    const { default: waitFor } = await (eval('import(`p-wait-for`)') as Promise<
```
artursudnik, line 44: Why is `p-wait-for` necessary to be imported like this?

JGiter: It is problem we've started dealing with for some time. A lot of npm packages are migrated to ESM-only. And the only way to import ESM from CommonJs (which did-ipfs-store is) is dynamic import https://www.typescriptlang.org/docs/handbook/esm-node.html. We also wrapping `import` in `eval` to prevent typescript from changing `import` on `require`

artursudnik: oh, so this is it! thx for explaining. :)

packages/did-ipfs-store/src/didStore.ts
jrhender: @JGiter yes, makes sense. In principle, could we return while the status was still `pinning`? I'm not saying that's better necessarily, but just curious what the pros/cons are in your mind.

JGiter: I don't see pros of returning when data is still pinning. User will still need to wait for the data to become available. Thus task of awaiting will fall on him

packages/did-ipfs-store/src/didStore.ts
jrhender, line 44: @JGiter will use of dynamic import cause issues with the jest (https://energyweb.atlassian.net/browse/SWTCH-1508?focusedCommentId=23518) ?

packages/did-ipfs-store/src/didStore.ts
```diff
@@ -48,6 +67,36 @@ export class DidStore implements IDidStore {
     }
   }
 
+  /**
+   * Checks if file is pinned on cluster
+   * @param cid CID
+   * @param isFullyReplicated specifies whether to check if data is fully replicated or pinned at least on one node
```
jrhender, line 73: - Is there a different in meaning between "pinned" and "replicated" in the IPFS context? Maybe there is, so best to be precise and use "pinned" rather than "replicated"
- Seems to check REPLICATION_FACTOR_MIN rather than "one node"?
```suggestion
   * @param isFullyReplicated specifies whether to check if data is pinned to all nodes in the cluster or pinned to at least REPLICATION_FACTOR_MIN nodes
```

packages/did-ipfs-store/src/didStore.ts
jrhender: > User will still need to wait for the data to become available. Thus task of awaiting will fall on him

They might not need to await for the data to be available if they aren't trying to read the data. If they are just writing, then they don't need to wait for read availability.

> Thus task of awaiting will fall on him

I think this could actually be an advantage, as `waitFor`'s behaviour might not suit every client. Currently, there is a `pinTimeout` command, but what if clients want to control other params of the `waitFor` behaviour (e.g. polling frequency)

How about this as a solution? We keep the `save` method as is, but make the `add` method public so that a client could use it can have more control over the wait behaviour if they want to?

packages/did-ipfs-store/src/didStore.ts
```diff
@@ -48,6 +67,36 @@ export class DidStore implements IDidStore {
...
+   */
+  async isPinned(cid: string, isfullyReplicated = true): Promise<boolean> {
+    const repication_factor_min = isfullyReplicated
+      ? (await this.allocations(cid)).replication_factor_min
+      : REPLICATION_FACTOR_MIN;
```
jrhender, line 78: - Is this `clusterNodeCount` variable name that I've suggested accurate @JGiter ?
- why the snake case 🐍  on `repication_factor_min`?
- `repication_factor_min` is a confusing variable name because it is the same as `REPLICATION_FACTOR_MIN` but lowercase
```suggestion
    const clusterNodeCount = (await this.allocations(cid)).replication_factor_min;
    const repication_factor_min = isfullyReplicated
      ? clusterNodeCount
      : REPLICATION_FACTOR_MIN;
```

jrhender (commented):
Some further comments

packages/did-ipfs-store/src/didStore.ts
JGiter: Ah, right. Will try `wait-on` instead

packages/did-ipfs-store/src/didStore.ts
JGiter: The question is what is default behaviour? To wait or not to wait? I beleive that by default we should wait. The drawback is that those who doesn't need after save availability has to provide empty `waitForPinned`. But the drowback for the case when default `waitForPinned` is empty is bigger, because whoever need claim to be avalilable right after `save` will need to put significant efforts for `waitForPinned` implementation. Also it might be confusing for people that `save` doesn't saves

packages/did-ipfs-store/src/didStore.ts
JGiter: Ah, yeah it is a problem. I will have to implement it then

packages/did-ipfs-store/src/didStore.ts
JGiter: snake case variables are returned from cluster
`clusterNodeCount` isn't accurate either. Let it be just `nodes`
Agree on confusing `REPLICATION_FACTOR_MIN`. `isfullyReplicated` as I can see now is also not precise. Probably the best in this case is to mirror cluster configuration option by `replication` option which takes values from {LOCAL, MIN, MAX} enum

packages/did-ipfs-store/src/didStore.ts
```diff
@@ -48,6 +68,42 @@ export class DidStore implements IDidStore {
...
 
+  /**
+   * Checks if file is pinned on cluster
+   * @param cid CID
+   * @param replication specifies on how many nodes data should be pinned https://ipfscluster.io/documentation/guides/pinning/#replication-factors
+   */
+  async isPinned(cid: string, replication = REPLICATION.MIN): Promise<boolean> {
+    const { replication_factor_min, replication_factor_max } =
+      await this.allocations(cid);
+    const nodes =
```
artursudnik, line 79: a little note: at first sight it is unclear for me if this constant contains list of nodes or number of nodes. I needed to analyze the code to figure it out.

JGiter: That's right. I coundn't figured clear name for number of nodes expected to pin data on. How about `expectedReplication`?

artursudnik: or: `nodesCountToBeReplicatedTo` if it is not too long

nichonien: How about nodeReplicationCount?

JGiter: the same dwawback - node and count are part of replication

artursudnik: maybe adding explicit type to this constant will help understand this?
```typescript
const expectedReplication: number =
```

JGiter: for me its the same as adding `count` to `replication`

JGiter: in order to make the difference from the process more obvious, the name can be lengthened like `expectedReplicationFactor`

nichonien (dismissed):

artursudnik (approved):

PR #457, opened 2022-10-20
check pin status of not saved claim

JGiter:
### Summary

|             |   |
|-------------|---|
| Description | When there are no allocations of claim, then `isPinned` will return `false` instead of throwing 404 |
| Jira issue  | |

### DIFF
--- a/packages/did-ipfs-store/src/didStore.ts
+++ b/packages/did-ipfs-store/src/didStore.ts
@@ -73,9 +73,20 @@ export class DidStore implements IDidStore {
    * @param cid CID
    * @param replicationFactor specifies on how many nodes data should be pinned https://ipfscluster.io/documentation/guides/pinning/#replication-factors
    */
-  async isPinned(cid: string, replicationFactor = REPLICATION.MIN): Promise<boolean> {
-    const { replication_factor_min, replication_factor_max } =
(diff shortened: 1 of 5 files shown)


nichonien (approved):

packages/did-ipfs-store/src/didStore.ts
```diff
@@ -73,9 +73,20 @@ export class DidStore implements IDidStore {
...
+  ): Promise<boolean> {
+    const response = await this.allocations(cid);
+    if (!response.ok) {
+      if (response.status === 404) {
+        return false;
+      } else {
+        throw new Error(response.statusText);
+      }
+    }
+    const allocations = (await response.json()) as PinResponse;
```
jrhender, line 88: Maybe it's better to have the return type of `this.allocations` be casted so that casting is closer data to avoid potential duplication of the casting?
```suggestion
    const response: { json: () => PinResponse } = await this.allocations(cid);
    if (!response.ok) {
      if (response.status === 404) {
        return false;
      } else {
        throw new Error(response.statusText);
      }
    }
    const allocations = await response.json();
```

jrhender (approved):
A small possible suggestion but it looks good 👍

PR #458, opened 2022-10-24
task/error to indicate not persisted claim

JGiter:
### Summary

|             |   |
|-------------|---|
| Description |  |
| Jira issue  |   |

### DIFF
--- a/packages/did-ipfs-store/src/didStore.ts
+++ b/packages/did-ipfs-store/src/didStore.ts
@@ -2,7 +2,7 @@ import { IDidStore } from '@ew-did-registry/did-store-interface';
 import fetch from '@web-std/fetch';
 import { FormData } from '@web-std/form-data';
 import { Blob } from '@web-std/file';
-import axios from 'axios';
+import axios, { AxiosError } from 'axios';
 import { values } from 'lodash';
 import {
   AddResponse,
@@ -13,6 +13,7 @@ import {
   REPLICATION,
 } from './ipfs.types';
 import { waitFor } from './utils';
+import { ContentNotFound } from './errorrs';
 
 /**
  * Implements decentralized storage in IPFS cluster. Storing data in cluster allows to provide required degree of data availability
@@ -31,7 +32,7 @@ export class DidStore implements IDidStore {
   constructor(
     private url: string,
     private headers: Record<string, string> = {}
-  ) {}
+  ) { }
 
   /**
    * @param claim stringified claim. Supported types of claim content are `string` and `object`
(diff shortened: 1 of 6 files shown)


packages/did-ipfs-store/src/didStore.ts
```diff
@@ -54,9 +55,19 @@ export class DidStore implements IDidStore {
...
-    });
+    let content: any;
+    try {
+      ({ data: content } = await axios.get(`${this.url}/ipfs/${cid}`, {
+        headers: this.headers,
+      }));
+    } catch (e) {
+      const { response } = <AxiosError>e;
+      if (response?.status === 504) {
+        throw new ContentNotFound(cid, response.statusText);
```
jrhender, line 66: @JGiter why does `504`, which indicates a Gateway Timeout, correspond to `ContentNotFound`?
```suggestion
      if (response?.status === 504) {
        throw new ContentNotFound(cid, response.statusText);
```

JGiter: My goal is to handle case when content isn't persisted and therefore response is equivalent to 404. But I haven't found a way to check this. All methods to locate or return content respond with 504 if no data found within given timeout. Sinse we are migrating to cluster consisting from comparatively not large amount of nodes, the 504 is actually means that requested data not in cluster, and not that IPFS don't have enough time to find it

packages/did-ipfs-store/src/errorrs.ts
```diff
@@ -0,0 +1,9 @@
+export class ContentNotFound extends Error {
+  constructor(cid: string, reason?: string) {
+    super(
+      `Contend with identifier ${cid} was not found in DID Store ${
```
jrhender, line 4: small typo
```suggestion
      `Content with identifier ${cid} was not found in DID Store ${
```

packages/did-ipfs-store/src/didStore.ts
jrhender: @JGiter Okay, I'm opening to adding this potentially, but I just want to understand it a little bit more. Let me try to paraphrase the behaviour you've observed to make sure I understand it correctly:

When requesting a CID from the EnergyWeb IPFS Cluster:
- if the data is available, then it is being returned
- if the data is not available, then after some time, a `504` is returned

Is the above correct ⬆️ ?

JGiter:
@jrhender 
> if the data is not available, then after some time, a 504 is returned

This is what I am currently observing. The problem is that IPFS documentation neither says nothing what `GET` returns when there is no data nor suggest a proper way to check if data exists

jrhender:
> This is what I am currently observing. The problem is that IPFS documentation neither says nothing what `GET` returns when there is no data nor suggest a proper way to check if data exists

Looking at the IPFS Gateway documentation it does seem like `504` might be the expected response when content is not found: https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#504-gateway-timeout

The comment in this question suggest that this is the case as well: https://stackoverflow.com/questions/71181813/504-gateway-time-out-openresty
> The ipfs.io gateway can't find the file in the DHT, try another gateway and make sure the node with the files pinned is reachable.

So maybe this is the right approach 👍

packages/did-ipfs-store/src/didStore.ts
```diff
@@ -54,9 +55,19 @@ export class DidStore implements IDidStore {
...
-      headers: this.headers,
-    });
+    let content: any;
+    try {
+      ({ data: content } = await axios.get(`${this.url}/ipfs/${cid}`, {
+        headers: this.headers,
+      }));
+    } catch (e) {
+      const { response } = <AxiosError>e;
+      if (response?.status === 504) {
```
jrhender, line 65: what do you think of adding this comment to explain why this check/conversion?
```suggestion
      // 504 is the expected response code when IPFS gateway is unable to provide content within time limit
      // https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#504-gateway-timeout
      // In other words, this is the expected response code if the traversal of the DHT fails to find the content
      if (response?.status === 504) {
```

jrhender (commented):
Looks good, just a small suggestion for documentation to explain for future readers

jrhender (approved):
Looks good, thanks @JGiter

PR #459, opened 2022-10-28
bug/SWTC-1459 missing revert data

JGiter:
### Summary

|             |   |
|-------------|---|
| Description | In order to send transaction to offerable identity two conditons must be met. Signer should own offerable identity and offerable identity should own `did` being updated. First is checked in OfferableIdentity.sol. This PR adds second|

Also had to pin @sphereon/pex, because recent bugfix introduced breaking changes

### DIFF
--- a/packages/credentials-interface/package.json
+++ b/packages/credentials-interface/package.json
@@ -47,7 +47,7 @@
     "url": "git+https://github.com/energywebfoundation/ew-did-registry.git"
   },
   "dependencies": {
-    "@sphereon/pex": "^1.1.0",
+    "@sphereon/pex": "1.1.3",
     "@types/lodash": "^4.14.181",
     "joi": "^17.6.0",
     "lodash": "^4.17.21"
--- a/packages/proxyIdentity/src/offerableIdentityOperator.ts
+++ b/packages/proxyIdentity/src/offerableIdentityOperator.ts
@@ -14,6 +14,7 @@ import {
 } from '@ew-did-registry/did-ethr-resolver';
 import { abi as identityAbi } from '../build/contracts/OfferableIdentity.json';
 import { abi as erc1056Abi } from '../constants/ERC1056.json';
+import { OnlyOwnerAllowed } from './errors';
 
 const { Interface, formatBytes32String } = utils;
 const { PublicKey, ServicePoint } = DIDAttribute;
@@ -28,7 +29,7 @@ export class OfferableIdenitytOperator extends Operator {
    * @param identityAddr - Address of controlled offerable identity
    */
   constructor(
-    owner: EwSigner,
+    private owner: EwSigner,
     settings: RegistrySettings,
     identityAddr: string
   ) {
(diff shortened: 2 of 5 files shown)


jrhender (commented):
How hard would it be to add a test that covers this scenario?

packages/proxyIdentity/src/offerableIdentityOperator.ts
```diff
@@ -77,6 +78,10 @@ export class OfferableIdenitytOperator extends Operator {
     validity?: number
   ): Promise<BigNumber> {
     const identity = addressOf(did);
+    const signer = await this.getAddress();
+    if (!((await this.identityOwner(identity)) !== signer)) {
```
nichonien, line 82:
```suggestion
    if (!((await this.identityOwner(identity)) === signer)) {
```

nichonien (approved):

packages/proxyIdentity/src/offerableIdentityOperator.ts
```diff
@@ -77,6 +83,10 @@ export class OfferableIdenitytOperator extends Operator {
     validity?: number
   ): Promise<BigNumber> {
     const identity = addressOf(did);
+    const identityOwner = await this.identityOwner(did);
+    if (!(identityOwner === this.identity.address)) {
+      throw new OnlyOwnerAllowed(identityOwner, await this.getOwnerAddress());
```
jrhender, line 88: The error message is `Signer ${signer} is not allowed to update DID identity ${identity}`, so shouldn't `did` be the second param?
```suggestion
      throw new OnlyOwnerAllowed(identityOwner, did);
```

JGiter: Do you mean that order of params in message should be the same as order in error call? I put identity on the first place in constructor, because `identity` seems to me primary in this event. So that when you write `OnlyOwnerAllowed` the first what you expect is whose owner? 
Do you think preserving order has some benefits

packages/credentials-interface/package.json
```diff
@@ -47,7 +47,7 @@
     "url": "git+https://github.com/energywebfoundation/ew-did-registry.git"
   },
   "dependencies": {
-    "@sphereon/pex": "^1.1.0",
+    "@sphereon/pex": "1.1.3",
```
jrhender, line 50: Is this change relevant to this PR? It's a different package, so I think it should be removed

jrhender (changes requested):
Just an unrelated change to `credentials-interface` that I think should be removed

packages/credentials-interface/package.json
JGiter: `@sphereon/pex` introduced breaking changes in 1.1.4

jrhender: Ok, thanks for the explanation. This maybe indeed be a good change to make. Still, I think it is separate to this PR, so would be best made in a separate PR, so that we can more easily revert/isolate the change if necessary. What do you think?

JGiter: Ok, make sence https://github.com/energywebfoundation/ew-did-registry/pull/460

jrhender (approved):
Thanks, looks good

PR #462, opened 2022-11-09
task/switch-1683 update contract deployment script

nichonien:
### Summary

|             |   |
|-------------|---|
| Description | Update truffle config and contract migration script for deployment  |
| Jira issue  |  https://energyweb.atlassian.net/browse/SWTCH-1683  |


JGiter (commented):
Since we are going to deploy with `truffle migrate` in proxyidentity can we remove https://github.com/energywebfoundation/ew-did-registry/blob/development/scripts/deployIdentityManager.js?

nichonien:
> Since we are going to deploy with `truffle migrate` in proxyidentity can we remove https://github.com/energywebfoundation/ew-did-registry/blob/development/scripts/deployIdentityManager.js?

Hmm, nice point. I will remove it.

JGiter (approved):

nichonien:
@jrhender @whitneypurdum I see the `snyk` test failing but I cannot know why, it says I don't have access though I was able to see the results in my previous PR.

packages/proxyIdentity/migrations/2_identityManager_migrations.js
```diff
@@ -2,10 +2,13 @@
 const { deployProxy } = require('@openzeppelin/truffle-upgrades');
 
 // needs to be updated with actual address
-const libraryAddress = '0x4B0897b0513fdC7C541B6d9D7E929C4e5364D2dB';
+const libraryAddress = '0x4c392f00a17B1b77aB059E6dc9ccE91F74113c66';
```
jrhender, line 5: @nichonien is this comment of `// needs to be updated with actual address` still accurate? What contract is `'0x4c392f00a17B1b77aB059E6dc9ccE91F74113c66'`?

nichonien: That is correct address for the latest deployment. This comment is from the standpoint when someone else uses this script, this address would be irrelevant at that point of time for a fresh deployment. Maybe I should rephrase this comment?

nichonien: Updated the comment.

jrhender (approved):
The failing Snyk check for licenses, which I think we still need to adjust. So this is good from my perspective.

PR #464, opened 2022-12-07
fix: upgrade dependencies

nichonien:


jrhender (approved):
Looks good to me, thanks @nichonien

PR #465, opened 2023-01-06
feat: validate duplicate service endpoint addition

nichonien:
### Summary

|             |   |
|-------------|---|
| Description | Add validation check for duplicate service endpoint addition in services for a DID  |
| Jira issue  |  https://energyweb.atlassian.net/browse/SWTCH-1822  |

### DIFF
--- a/packages/did-ethr-resolver/src/implementations/operator.ts
+++ b/packages/did-ethr-resolver/src/implementations/operator.ts
@@ -150,8 +150,16 @@ export class Operator extends Resolver implements IOperator {
     if (validity < 0) {
       throw new Error('Validity must be non negative value');
     }
-    if (didAttribute === ServicePoint && !updateData.value?.serviceEndpoint) {
-      throw new Error('Service Endpoint is required');
+    if (didAttribute === ServicePoint) {
+      if (!updateData.value?.serviceEndpoint) {
+        throw new Error('Service Endpoint is required');
+      }
+      const userDIDDoc = await this.read(did);
+      for (const svc of userDIDDoc.service) {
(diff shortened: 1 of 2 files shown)


JGiter (commented):
Are you sure this will prevent duplication? For me it looks like this change prevents updating of existing service. I think duplication is caused by reading of logs

nichonien:
> Are you sure this will prevent duplication? For me it looks like this change prevents updating of existing service. I think duplication is caused by reading of logs

For every DID Document update we only add new service endpoint for a credential. So this would block any change for which the service endpoint exists, be it new addition or updation.

jrhender (commented):
The code change looks good to me.
However, I recommend that a test be added for this duplicate service case. I think a test added to this file would do the trick: https://github.com/energywebfoundation/ew-did-registry/blob/development/packages/did-ethr-resolver/test/did-operator.test.ts

JGiter (commented):
Will service be duplicated if someone updates it not by ew-did-registy, but directly calling EthrDidRegistry?

nichonien:
> Will service be duplicated if someone updates it not by ew-did-registy, but directly calling EthrDidRegistry?

Yeah, the only way to prevent this is to update the didRegistry contract which isn't feasible.

nichonien:
> > Yeah, the only way to prevent this is to update the didRegistry contract which isn't feasible.
> 
> Do you think it is correct behavior of `Resolver.read(did)`, when it is returns duplicate services?

Aah, that's a nice point. Although IMO I would keep `Resolver.read(did)` as it is since the addition of duplicate check for service endpoint will prevent future occurrences and the resolver remains generic and returns all services without any filter.

jrhender (approved):

PR #476, opened 2023-06-15
Update @sphereon/pex@2.0.1

JGiter:


packages/did-ipfs-store/test/did-store.test.ts
```diff
@@ -20,15 +20,15 @@ const testSuite = function () {
     const infuraCid = cidPinnedInInfura;
     const claim = await this.gateway.get(infuraCid);
     const clusterCid = await this.cluster.save(claim);
-    expect(await this.cluster.get(infuraCid, 120000)).equal(claim);
-    expect(await this.cluster.get(clusterCid)).equal(claim);
+    expect(await this.cluster.get(infuraCid)).equal(claim);
+    expect(await this.cluster.get(clusterCid, 60_000)).equal(claim);
```
JGiter, line 24: resolving by cluster CID takes more time then by Infura CID

packages/did-ipfs-store/test/did-store.test.ts
```diff
@@ -20,15 +20,15 @@ const testSuite = function () {
...
   });
 
   it('get() should throw error if data was not persisted', async function () {
     const cid = await Hash.of(chance.string());
     return Promise.all([
-      expect(this.cluster.get(cid)).to.be.rejectedWith(ContentNotFound),
-      expect(this.gateway.get(cid)).to.be.rejectedWith(ContentNotFound),
+      expect(this.cluster.get(cid, 1_000)).to.be.rejectedWith(ContentNotFound),
```
JGiter, line 30: somehow long duration of resolving non existing claim affects other tests. This small timeout should not invalidate this test, because resolution by Infura CID takes no more than 600ms

package-lock.json
JGiter, line 1: changed package-lock.json is side effect of 
```
lerna add @sphereon/pex
lern setup
```
which looks like a more proper way to update, then calling `npm i right in credentials-interface`

packages/credentials-interface/package.json
```diff
@@ -47,7 +47,9 @@
     "url": "git+https://github.com/energywebfoundation/ew-did-registry.git"
   },
   "dependencies": {
-    "@sphereon/pex": "1.1.3",
+    "@sphereon/pex": "^2.0.1",
+    "@sphereon/pex-models": "^2.0.2",
```
jrhender, line 51: I don't see `@sphereon/pex-models` imported anywhere (though maybe I just missed it 😅 ). Does it need to be a dependency?

JGiter: Yes, we are importing presentation definitions there https://github.com/energywebfoundation/ew-did-registry/blob/e46e2edc7ecd18f14e30b2b10527fff6b737a17f/packages/credentials-interface/src/exchange.utils.ts#L6.

jrhender: Ah okay, thank you

jrhender (approved):
Looks good to me, thanks @JGiter

PR #480, opened 2023-06-22
Make Credential.id required

JGiter:
In `@sphereon/ssi-types` Credential.id is optional https://github.com/Sphereon-Opensource/ssi-sdk/blob/8f720df1ef84571ea2ef6eac1e3cd706849e53bc/packages/ssi-types/src/types/vc.ts#L21, but in all ours credential it is required

### DIFF
--- a/packages/credentials-interface/src/credentials.types.ts
+++ b/packages/credentials-interface/src/credentials.types.ts
@@ -21,6 +21,7 @@ export enum CredentialType {
   EWFRole = 'EWFRole',
 }
 export interface Credential<T extends CredentialSubject> extends ICredential {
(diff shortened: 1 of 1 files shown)


jrhender:
IMO, I won't merge this because the id property is optional according to VC data model spec: https://www.w3.org/TR/vc-data-model/#identifiers
> If the id [property](https://www.w3.org/TR/vc-data-model/#dfn-property) is present:

However, if you would like to merge @JGiter , I think it is okay 👍

jrhender (approved):
After discussion in standup today, this change seems reasonable based on id being required across EW apps

PR #483, opened 2023-06-29
Skip lib types check

JGiter:


jrhender (approved):
Thanks @JGiter .
The PR could be merged as is, but I suggest that you amend you commit with a more verbose description so that someone looking at the history could know why this change was made.
Paraphrasing the rationale you provided elsewhere, I suggest something like this:
> skipLibCheck is justified when several libraries provides inconsistent definition of the same type, which is the case currently
(see https://github.com/energywebfoundation/ew-did-registry/actions/runs/5411227314/jobs/9833795871 and
https://www.typescriptlang.org/tsconfig#skipLibCheck).
We have several versions of types/node as well as other packages, not all of which are fixable.
For example @web-std/blob and @web-std/fetch require different versions of @web-std/stream, while @web-std/blob required by @web-std/fetch

nichonien (approved):

PR #580, opened 2026-09-08
fix(jwt): issue `iat` in seconds per RFC 7519

### LINKED ISSUE #579: jwt: `iat` is issued in milliseconds, violating RFC 7519 NumericDate
## Summary

`JWT.sign()` in `@ew-did-registry/jwt` sets the `iat` claim with
`new Date().getTime()` — **milliseconds**. Per RFC 7519 (§4.1.6 and the
NumericDate definition in §2) `iat` must be **seconds** since the epoch, which is
also how the same method already emits `exp`
(`Math.trunc(expirationTimestamp / 1000)`).

## Impact

- Tokens are not RFC 7519 compliant. Standard verifiers that read `iat`
  (e.g. `jsonwebtoken`'s `maxAge` freshness check) treat the token as issued
  ~50,000 years in the future, so age-based checks never trip.
- `iat` and `exp` use different units **within the same token**.
- `iam-client-lib`'s `createDelegateProof()` builds one payload two ways: the
  `EIP191` branch calls this library (ms `iat`), the `ES256` branch calls
  `jsonwebtoken.sign` (spec-correct seconds `iat`) — inconsistent output from a
  single function.

The existing test suite doesn't catch it: every case passes `noTimestamp: true`,
and `JwtBase.verifyEIP191` never inspects `iat`.

## Reproduction

Published packages, no chain required:

```js
const { JWT } = require('@ew-did-registry/jwt');
const { Keys } = require('@ew-did-registry/keys');
const jsonwebtoken = require('jsonwebtoken');

(async () => {
  const jwt = new JWT(new Keys());

  const token = await jwt.sign({ claim: 'x' });
  console.log('iat        =', jwt.decode(token).iat);            // 1788886121928  (ms)
  console.log('now seconds =', Math.floor(Date.now() / 1000));   // 1788886121

  console.log('jsonwebtoken iat =', jsonwebtoken.decode(
    jsonwebtoken.sign({ claim: 'x' }, 'secret')).iat);           // 1788886121  (seconds)

  const t2 = await jwt.sign({ claim: 'x' }, { expirationTimestamp: Date.now() + 60_000 });
  const d2 = jwt.decode(t2);
  console.log('exp - iat  =', d2.exp - d2.iat);                  // -1787097235782  (should be ~60)
})();
```

## Fix

Emit `Math.floor(Date.now() / 1000)`. PR:

SafraNako:
### Summary

|             |   |
|-------------|---|
| Description | `JWT.sign` emitted the `iat` claim with `Date.now()` (**milliseconds**) — ~1000x the RFC 7519 NumericDate value, and inconsistent with `exp` (seconds) in the same token and with tokens issued via `jsonwebtoken` (e.g. the `ES256` branch of `iam-client-lib`'s `createDelegateProof`). Now emits seconds; adds regression tests. |
| Jira issue  | n/a — external contribution |

Closes #579

#### Type of request

### List of Features

- `iat` is now a NumericDate in **seconds** since the epoch (RFC 7519 §4.1.6 / §2), matching how `exp` is already emitted
- new `[JWT iat/exp timestamps]` block in `packages/jwt/test/jwt.test.ts`: `iat` is integer seconds ≈ now; `exp - iat` ≈ 60 (same unit); `iat` omitted under `noTimestamp`

### Notes

- One-line behaviour change in `packages/jwt/src/JWTEIP191.ts`; nothing in this repo or `iam-client-lib` reads `iat` as milliseconds (`grep -rn '\biat\b'` → only the type declaration and this assignment), so the change is safe.
- Verified against the patched build: `iat` is now seconds, `exp - iat === 60`, `noTimestamp` still omits `iat`, and the sign/verify round-trip is unaffected.

### DIFF
--- a/packages/jwt/src/JWTEIP191.ts
+++ b/packages/jwt/src/JWTEIP191.ts
@@ -66,7 +66,10 @@ export class JWT extends JwtBase implements IJWT {
       payload.sub = subject;
     }
     if (!noTimestamp) {
-      payload.iat = new Date().getTime();
+      // `iat` is a NumericDate: seconds since the epoch (RFC 7519, 4.1.6 / 2).
+      // `Date.now()` returns milliseconds, which is ~1000x too large and
+      // inconsistent with `exp` below (and with `jsonwebtoken`-issued tokens).
+      payload.iat = Math.floor(Date.now() / 1000);
     }
     if (expirationTimestamp) {
       if (expirationTimestamp < Date.now()) {
--- a/packages/jwt/test/jwt.test.ts
+++ b/packages/jwt/test/jwt.test.ts
@@ -104,3 +104,39 @@ describe('[JWT PACKAGE]', () => {
     });
   });
 });
+
+describe('[JWT iat/exp timestamps]', () => {
+  const signer = new JWT(new Keys());
+  // `sign` mutates its payload argument, so each test uses a fresh object
+  const freshPayload = () => ({ claim: 'test' });
+
+  it('`iat` should be a NumericDate in seconds, not milliseconds (RFC 7519)', async () => {
+    const nowSeconds = Math.floor(Date.now() / 1000);
+    const token = await signer.sign(freshPayload());
+    const { iat } = signer.decode(token) as { iat: number };
+
+    expect(iat).to.be.a('number');
+    expect(Number.isInteger(iat)).to.equal(true);
+    // within a 5s window of "now" expressed in seconds; a milliseconds value
+    // would be ~1000x larger and fail this by many orders of magnitude
+    expect(Math.abs(iat - nowSeconds)).to.be.lessThan(5);
+  });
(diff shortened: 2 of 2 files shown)


SafraNako:
Hi maintainers 👋

Just a gentle follow-up, this PR has been open for about 10 days now. It's a small, self-contained fix: `payload.iat` in `JWTEIP191.ts` was being set with `Date.getTime()` (milliseconds) instead of seconds, which violates [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.6) `NumericDate` semantics and could make `exp`/token-expiry checks behave unexpectedly downstream. Tests are included.

No rush at all, just wanted to make sure it didn't slip through unnoticed. Happy to adjust anything if you'd like a different approach. Thanks for maintaining this project!

